PR #18, opened 2011-03-16
Core view convert null to array

henrikbjorn:
Makes it easy to create actions that are just used to render a static template.

    function aboutAction() {
    }

Maybe this is too much magic but reduces redundant `return array();` code.


javiereguiluz:
I think it's safe to close this pull request as "won't fix" because we now have an official way to render static templates without a controller.

PR #115, opened 2012-04-20
Phpcr converter

lsmith77:


fabpot:
I think it comes to usage and maintenance. I'm not a user of PHPCR and usage is (tell me if I'm wrong) very low compared to the ORM which is almost standard. So, I'm closing this now.

lsmith77:
sure

PR #196, opened 2013-02-13
[DoctrineParamConverter] Added option to specify not found error message

apfelbox:
The error message can now be added to the options array in the ParamConverter definition with the key `"notFoundError"`

Fixes #195

### DIFF
--- a/Request/ParamConverter/DoctrineParamConverter.php
+++ b/Request/ParamConverter/DoctrineParamConverter.php
@@ -52,7 +52,8 @@ class DoctrineParamConverter implements ParamConverterInterface
         }
 
         if (null === $object && false === $configuration->isOptional()) {
-            throw new NotFoundHttpException(sprintf('%s object not found.', $class));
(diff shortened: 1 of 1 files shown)


ghanbari:
@fabpot 
I use fos restbundle, that permit use exception message, without use of this capability of rest bundle, i must render template files for rest requests that is not good

ghanbari:
@fabpot 
I use multiply `@ParamConvertor` on my method and i want show a appropriate message, but i can not detect in `404.json.twig`, which object wasn't found,

also i use **fos rest bundle** that permit to me show exception message to user, but i no want show class name to final user, (that is a good ability, if and if i can customize exception message)

baptistedonaux:
Maybe Sensio should review its position on this issue. Allow custom message can simplify the error management for many developers. @apfelbox @fabpot

delboy1978uk:
> Closing as exception messages are not meant to be displayed to the user.

What if it is an API client? And indeed, if we shouldn't see the exception messages, we should be able to replace them in the annotations, I think it is worth adding this nice feature

PR #213, opened 2013-04-02
[Template] Added lowercaseName option

getme:
Can we merge this? here''s summary

Often controller actions named as `viewSummaryAction()` or `viewSomething()`
but desired twig template names are `view_summary.html.twig` and `view_something.html.twig` but with current @Template it seems not possible to achieve this (except full template path like @Template(FooBarBundle:Controller:view_summary.html.twig))

so, `@Template(lowercaseName=true)` will do all work above, I think idea is clear.
Thanks

### DIFF
--- a/Configuration/Template.php
+++ b/Configuration/Template.php
@@ -50,6 +50,13 @@ class Template extends ConfigurationAnnotation
     protected $streamable = false;
 
     /**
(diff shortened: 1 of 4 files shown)


fabpot:
see #507

PR #226, opened 2013-06-26
Possibility to change default templating engine.

rafalwrzeszcz:
For people who don't use **Twig** ;).

### DIFF
--- a/Configuration/Template.php
+++ b/Configuration/Template.php
@@ -29,11 +29,11 @@ class Template extends ConfigurationAnnotation
     protected $template;
 
     /**
-     * The template engine used when a specific template isnt specified
+     * The template engine to use
      *
      * @var string
      */
-    protected $engine = 'twig';
+    protected $engine;
 
     /**
      * The associative array of template variables.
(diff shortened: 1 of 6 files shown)


kralos:
@rafalwrzeszcz Just stumbled across this, you can switch engines using a response listener.

```php
<?php

namespace AppBundle\Http\Response;

use Symfony\Component\HttpKernel\Event\GetResponseForControllerResultEvent;
use Symfony\Bundle\FrameworkBundle\Templating\TemplateReference;

class CsvResponseListener
{
    /**
     * Use php templating for csv responses so we can utilize fputcsv etc
     * Template would look like:
     *     $output = fopen('php://output', 'w');
     *     fputcsv($output, ['a', 'b', 'c']);
     *     fputcsv($output, ['1', '2', '3']);
     *     fclose($output);
     *
     * @param \Symfony\Component\HttpKernel\Event\GetResponseForControllerResultEvent $event
     */
    public function onKernelView(GetResponseForControllerResultEvent $event)
    {
        $request = $event->getRequest();

        $format = $request->getRequestFormat();
        if ($format !== 'csv') {
            return;
        }

        $template = $request->attributes->get('_template');
        if (!$template instanceof TemplateReference) {
            return;
        }

        $template->set('engine', 'php');
    }
}
```

Then register your listener:
```yaml
---
services:
    app.http.response.csv_response_listener:
        class: AppBundle\Http\Response\CsvResponseListener
        tags:
            - { name: kernel.event_listener, event: kernel.view, method: onKernelView, priority: 100 }
```

PR #253, opened 2013-12-04
[WIP] Routing template

ewgRa:
There is patch for ability to set template from routing, like this:

```
pageAwesome:
    path:      /{product}-awesome
    defaults:  { _controller: Bundle:Product:show, _template: Bundle:Product:show-awesome.html.twig}

page:
    path:      /{product}
    defaults:  { _controller: Bundle:Product:show}
```

Use case: I want reuse my controller code for different views, and don't want have if or switch logic and view arguments in controller.
Can be used with Template annotation, or without.

What do you think?

### DIFF
--- a/Configuration/Method.php
+++ b/Configuration/Method.php
@@ -68,7 +68,7 @@ class Method extends ConfigurationAnnotation
     }
 
     /**
-     * Only one cache directive is allowed
+     * Only one method directive is allowed
      *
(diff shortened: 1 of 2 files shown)


lemoinem:
If you are still interested in the feature, you can find it in https://github.com/wemakecustom/TemplateInRouteBundle

PR #266, opened 2014-01-07
Specifying templates directly in Routes

lemoinem:
This allows using a `_template` attribute in the route directly to specify a template, just as if a `@Template` annotation was used.

Disclaimer: This PR is basically a reimplementation of sensiolabs/SensioFrameworkExtraBundle#253 I tried to reach the initial author of the PR (@ewgRa), but to no avail.

I re-implemented the PR because my solution is much simpler (8 lines) and avoid declaring a new EventSubscriber.
I also took the occasion to write some documentation for it.

If specified through the route, the `_template` parameter will be removed from `_route_params`.

This PR contains an additional feature with respect to the original one: If `true` is specified as the value for `_template`, then the TemplateListener will use the Template auto guessing. I though this might be useful when Routes are generated using PHP or another configuration loader supporting literals.
This allow us to avoid having to insert an additional `use Template` and `@Template` annotation in very simple cases.

Moreover, if an invalid value is specified for `_template`, the listener will simply ignore it (as opposed to crashing before).

### DIFF
--- a/EventListener/TemplateListener.php
+++ b/EventListener/TemplateListener.php
@@ -74,7 +74,7 @@ class TemplateListener implements EventSubscriberInterface
         }
 
         if (isset($routeParams['_template'])) {
-            if(is_string($routeParams['_template'])) {
+            if (is_string($routeParams['_template'])) {
                 $configuration->setTemplate($routeParams['_template']);
             } elseif (true === $routeParams['_template']) {
                 $configuration->setTemplate(null);


lemoinem:
Thanks @fabpot for the feedback. I understand not every feature need to be included.

However, for the ones interested in the feature, you can find it in its own bundle: https://github.com/wemakecustom/TemplateInRouteBundle

PR #271, opened 2014-01-20
Repository method with arguments

tvlooy:
Example:

I have set this:

    /**
     * @ParamConverter(
     *    "post",
     *    class="AcmeBlogBundle:Post",
     *    options={
     *        "repository_method" = "findForYmdLink",
     *        "mapping": {"year": "year", "month": "month", "day": "day", "permalink": "permalink"}
     *    }
     * )
     */

My repository method looks like this:

    public function findForYmdLink($year, $month, $day, $permalink)
    {
        $creationDate = \DateTime::createFromFormat('Ymd', $year.$month.$day);

        return $this->findOneBy(array(
            'creationDate' => $creationDate,
            'permalink' => $permalink,
            'published' => true,
        ));
    }

My Post entity doesn't have the fields day, month and year but has creationDate instead (DateTime). The current implementation of the converter doesn't pass the 3 fields because they are not in the metadata. I don't see why they should be if I pass a custom repository_method. This patch changes that behaviour.

With the current implementation, I would also have to change the repository method to this:

    public function findForYmdLink($params)
    {
        list($year, $month, $day, $permalink) = $params;
        ...

This patch changes that behaviour by calling invokeArgs on the reflectionMethod so the exact parameters are in the method signature.

### DIFF
--- a/Request/ParamConverter/DoctrineParamConverter.php
+++ b/Request/ParamConverter/DoctrineParamConverter.php
@@ -90,10 +90,12 @@ class DoctrineParamConverter implements ParamConverterInterface
         }
 
         try {
-            return $this->getManager($options['entity_manager'], $class)->getRepository($class)->$method($id);
+            $result = $this->getManager($options['entity_manager'], $class)->getRepository($class)->$method($id);
         } catch (NoResultException $e) {
-            return null;
+            $result = null;
         }
+
+        return $result;
     }
 
     protected function getIdentifier(Request $request, $options, $name)
@@ -154,8 +156,7 @@ class DoctrineParamConverter implements ParamConverterInterface
         foreach ($options['mapping'] as $attribute => $field) {
             if ($metadata->hasField($field)
                 || ($metadata->hasAssociation($field) && $metadata->isSingleValuedAssociation($field))
-                || $mapMethodSignature)
-            {
+                || $mapMethodSignature) {
                 $criteria[$field] = $request->attributes->get($attribute);
             }
         }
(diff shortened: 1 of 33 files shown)


tvlooy:
@fabpot just merged master, resolved conflicts. All tests pass

Resources/doc/annotations/converters.rst
```diff
@@ -177,6 +177,40 @@ to add joins to the query), you can add the ``repository_method`` option::
...
+    {
+        public function findByFullName($firstName, $lastName)
+        {
+            ...
+        }
+    }
+
+.. tip::
+
+   With de default implementation, the ``firstName`` and ``lastName``
```
fabpot: I would have said: `When map_method_signature is true, ...` because `default implementation` does not refer to anything.

Resources/doc/annotations/converters.rst
```diff
@@ -177,6 +177,40 @@ to add joins to the query), you can add the ``repository_method`` option::
     {
     }
 
+The specified repository method will be called with the criteria in an ``array``
+as parameter. This is a good fit with Doctrine's ``findBy`` and ``findOneBy``
+methods.
+
+There are cases where you want to you use your own repository method and you
+want to map the criteria to the method signature. This is possible when you set
+the ``map_method_signature`` option to true. The default is false.
```
fabpot: You need to finish the sentence by `::` instead of `.` to make the following block a code block.

Request/ParamConverter/DoctrineParamConverter.php
```diff
@@ -162,16 +170,42 @@ protected function findOneBy($class, Request $request, $options)
...
+            $result = null;
+        }
+
+        return $result;
+    }
+
+    private function findDataByMapMethodSignature($em, $class, $repositoryMethod, $criteria)
+    {
+        $repository = $em->getRepository($class);
+        $repositoryClass = get_class($repository);
```
fabpot: You can remove this and use `$repository` directly as an argument to the Reflection constructors.

fabpot:
@tvlooy Don't take into account my comments, I'm going to have a look at the running code first on my machine to see how it behaves as I doubt it works the way I want (but I'm probably wrong here). Coming back here later today.

Resources/doc/annotations/converters.rst
```diff
@@ -177,6 +177,40 @@ to add joins to the query), you can add the ``repository_method`` option::
...
+
+    /**
+     * @Route("/user/{first_name}/{last_name}")
+     * @ParamConverter("user", class="CtorsBlogBundle:User", options={
```
tvlooy: Should be Acme

fabpot:
@tvlooy I've merged your PR and slightly change the implementation to take care of optional arguments and make the error message a little different. You can see those changes in 1c3ed3; your original tests still pass.

tvlooy:
Thanks!

PR #322, opened 2014-08-25
Guess template name based on declared method name when using @Template

tomat:
Stumbled upon this while developing on OS X which has a case-insensitive filesystem. For various non-obvious reasons I ended up calling a controller action using `mycontrollerAction` instead of `myControllerAction`.

Since I was using the `@Template` annotation without any parameters, the template name was guessed, but the guess was based on what I called the controller as, not the actual controller method name, resulting in `mycontroller.html.twig` instead of `myController.html.twig`.

This of course worked fine locally, but failed when pushed to our staging server (which has a case-sensitive file system).

### DIFF
--- a/Templating/TemplateGuesser.php
+++ b/Templating/TemplateGuesser.php
@@ -77,6 +77,14 @@ class TemplateGuesser
             $bundleName = null;
         }
 
+        foreach (get_class_methods($className) as $method) {
+            if ($method !== $matchAction[0] && strtolower($method) === strtolower($matchAction[0])) {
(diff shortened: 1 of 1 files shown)


fabpot:
The logic is the right one: template names are derived from the controller action method name.

tomat:
Technically it is derived from the name it is called as; which may be different from the actual method name since PHP is case insensitive in that regard. This patch would make it actually check the declared method name instead of relying on the called name.

But I guess it could be argued that it is not at all supported to call methods with another case than the exact way they are defined.

PR #327, opened 2014-09-20
ParamConverter to inject the current user

adamquaile:
When creating testable controllers, it's often desirable to not use the `getUser` function in the base controller, as this relies on the whole service container. 

In this case, we might add a dependency on the `SecurityContext` to achieve the same result, replicating this code: 

    public function getUser()
    {
        if (!$this->container->has('security.context')) {
            throw new \LogicException('The SecurityBundle is not registered in your application.');
        }

        if (null === $token = $this->container->get('security.context')->getToken()) {
            return;
        }

        if (!is_object($user = $token->getUser())) {
            return;
        }

        return $user;
    }

This parameter converter streamlines the process and injects the current user into the controller method instead using the following format:

    /**
     * @ParamConverter(converter="current_user", name="parameterName")
     */
    public function viewInboxAction(User $currentUser)


Resources/doc/annotations/converters.rst
```diff
@@ -203,6 +203,43 @@ is accepted. You can be stricter with input given through the options::
...
+     public function sendMessageAction(User $currentUser, User $recipient)
+
+If you type-hint against the UserInterface, or an interface which extends UserInterface,
+you may omit this annotation entirely, like this:
+
+    use Symfony\Component\Security\Core\User\UserInterface;
+
+    // ...
+
+    public function myAccountAction(\Symfony\Component\Security\Core\User\UserInterface $currentUser)
```
fabpot: This can be simplified to `public function myAccountAction(UserInterface $currentUser)` as you have a `use` statement above.

Resources/config/converters.xml
```diff
@@ -28,5 +28,10 @@
         <service id="sensio_framework_extra.converter.datetime" class="%sensio_framework_extra.converter.datetime.class%">
             <tag name="request.param_converter" converter="datetime" />
         </service>
+
+        <service id="sensio_framework_extra.converter.current_user" class="Sensio\Bundle\FrameworkExtraBundle\Request\ParamConverter\CurrentUserParamConverter">
+            <tag name="request.param_converter" converter="current_user" />
+            <argument type="service" id="security.context" on-invalid="ignore" />
```
fabpot: should be `null` instead of `ignore`

Request/ParamConverter/CurrentUserParamConverter.php
```diff
@@ -0,0 +1,94 @@
...
+    {
+        if (null === $this->security) {
+            return false;
+        }
+
+        $userInterfaceClass = 'Symfony\Component\Security\Core\User\UserInterface';
+        $configuredClass = $configuration->getClass();
+        $interfaces = class_implements($configuredClass);
+
+        return in_array($userInterfaceClass, $interfaces) || ($configuredClass == $userInterfaceClass);
```
fabpot: parenthesis around the second condition can be removed.

Resources/doc/annotations/converters.rst
adamquaile: Done. 944ec39

Resources/config/converters.xml
adamquaile: Done. 944ec39

Request/ParamConverter/CurrentUserParamConverter.php
adamquaile: Done. 944ec39

adamquaile:
@fabpot Job failed on travis for transient composer network issue. Restarted and it's good again.

adamquaile:
@fabpot Updated from most recent master. Any other changes I should make?

mickaelandrieu:
nice addition :)

Request/ParamConverter/CurrentUserParamConverter.php
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     *
+     * @throws \LogicException When SecurityBundle not enabled, or badly configured
+     */
+    public function apply(Request $request, ParamConverter $configuration)
+    {
+        $param = $configuration->getName();
```
MacDada: @adamquaile `$param` is used after if+throw so the value could be fetched later (after if+throw)

Resources/doc/annotations/converters.rst
```diff
@@ -203,6 +203,33 @@ is accepted. You can be stricter with input given through the options::
...
+to be converted. These are both required in order to not conflict with the Doctrine
+ParamConverter.
+
+The example below illustrates applying this to one parameter, where the ``$recipient``
+could be handled by another ParamConverter.
+
+    /**
+     * @ParamConverter(converter="current_user", name="user")
+     */
+     public function sendMessageAction(User $currentUser, User $recipient)
```
apfelbox: @adamquaile a little late to the game, but wouldn't it suffice to just register your converter before the doctrine converter? The doctrine converter only touches scalars IIRC, so if your converter already transformed the value before, this should just work?

adamquaile: @apfelbox Yeah, it's been almost a year since opening this. I'm not sure if there's even a plan to merge this in at all. 

I think you're right, it would have solved that error, but the solution we ended up with a good approach which is more reliable and less reliant on priority/ordering.

iltar: You won't have this problem if you don't use your entity as session user

Request/ParamConverter/CurrentUserParamConverter.php
adamquaile: @MacDada Yes, you're right. I wouldn't mind making this change, if this PR is still wanted - it was opened a year ago!

Any maintainers know if there is a plan to merge this, or should we close it? @fabpot @jakzal

Resources/doc/annotations/converters.rst
apfelbox: @adamquaile I saw that after reading the new comment down below. I wanted to comment here though, as I think that this addition would still be useful today.

@iltar yeah, that is right. Nevertheless the session user is often times the user entity as this is really comfortable and provides practically no downsides (except of this example right here of course :wink:). Or does it have any other negative impact?

iltar: You also have the issue of the object changing during the session if the entity is change, merging it back, storing more information in there than required etc. I'm very happy I split those 2 when I wrote my security part as it saved me from a lot of headaches.

Request/ParamConverter/CurrentUserParamConverter.php
MacDada: @adamquaile Well, typehinting `UserInterface` / `EntityWithUserInterface` is exactly what I'm looking for, so I hope it's gonna be resolved soon :)

Request/ParamConverter/CurrentUserParamConverter.php
```diff
@@ -0,0 +1,94 @@
...
+    {
+        if (null === $this->security) {
+            return false;
+        }
+
+        $userInterfaceClass = 'Symfony\Component\Security\Core\User\UserInterface';
+        $configuredClass = $configuration->getClass();
+        $interfaces = class_implements($configuredClass);
+
+        return in_array($userInterfaceClass, $interfaces) || $configuredClass == $userInterfaceClass;
```
Ener-Getick: I'd prefer ``$configuredClass === $userInterfaceClass || is_subclass_of($configuredClass, $userInterfaceClass)`` which is simpler (and probably faster).

iltar:
@adamquaile I think this PR is no longer required. This functionality can be build either inside the SecurityBundle in symfony or in your app manually. This PR is actually the inspiration I've used to add a doc example: https://github.com/symfony/symfony-docs/pull/6438/files#diff-378394f006769e0d975a8e861369dc33R99

fabpot:
agreed

PR #336, opened 2014-12-05
Security voters reference in @security doc page

MatissJA:
Added more information about the security annotation and how to create new security voters.


fabpot:
Thank you @MatissJA.

PR #339, opened 2015-01-27
@Cache(public=false) is being ignored

wimvds:
I tried to override an action in a controller using the @Cache(public=false) annotation, only to find out it always used public for the Cache-Control header.

This PR fixes that issue, and a regression test was added as well.


fabpot:
Thank you @wimvds.

PR #340, opened 2015-01-30
Fix deprecation notices

fabpot:


EventListener/SecurityListener.php
```diff
@@ -79,7 +85,8 @@ private function getVariables(Request $request)
             'request' => $request,
             'roles' => array_map(function ($role) { return $role->getRole(); }, $roles),
             'trust_resolver' => $this->trustResolver,
-            'security_context' => $this->securityContext,
+            // needed for the is_granted expression function
+            'auth_checker' => $this->authChecker,
```
stof: you need to keep the variable named ``security_context`` to be compatible with older versions of the security expressions

stof:
you need to add a compiler pass to remove the injection of the ``security.context`` service when new services are available, to avoid triggering the deprecation notice of the SecurityContext

EventListener/SecurityListener.php
fabpot: no, it's only for internal usage.

stof: ah, I though the parent class in the Security component was also using this variable. But it indeed does not

fabpot:
@stof I've added the pass and some tests as well.

EventListener/SecurityListener.php
```diff
@@ -48,7 +54,7 @@ public function onKernelController(FilterControllerEvent $event)
             return;
         }
 
-        if (null === $this->securityContext || null === $this->trustResolver) {
+        if (null === $this->tokenStorage || null === $this->trustResolver) {
```
scaytrase: Should 
```php 
null === $this->tokenStorage
``` 
be something like
```php
(null === $this->securityContext && null === $this->tokenStorage)
```
?

```$tokenStorage``` will always be null for symfony prior 2.6, i think

fabpot: No, prior to 2.6, $tokenStorage is equivalent to $securityContext, see above.

PR #342, opened 2015-02-03
Test that a token is set to use the @Security tag

iamluc:
Fix a bug when the Security annotation is used on a controller which is not behind a firewall


fabpot:
Thank you @iamluc.

fabpot:
updated to the current bundle version in eaa31b6

Issue #345, opened 2015-03-18
BC break Cache annotation


Thorry84:
In 3.0.5 the following change was made with regards to the Cache annotation to fix the response being public or private. https://github.com/sensiolabs/SensioFrameworkExtraBundle/commit/4c2e3e5d03857cc13fcf2a0bced34c587f59677e

However when the public key isn't added it always sets the response as being private adding the cache-control: private. Before this change was made it would not set the cache-control header at all, which is the desired behavior (I think).

To fix this I think the code should check whether the value was set in the first place and not add a header when it wasn't set.

fabpot:
@Thorry84 Can you propose a pull request to fix this regression? Thanks.

Thorry84:
Done!

PR #348, opened 2015-03-27
Fix BC break issue #345 cache public

Thorry84:
Fixes issue #345

When cache is not set to public nor set to private the response should
not be marked as public nor private


fabpot:
Thank you @Thorry84.

PR #349, opened 2015-04-01
[Bugfix] Vary header should not be overridden

Nyholm:
We should not overwrite the vary header of the response if we have not configured vary-header in our cache annotation. 

The issue is because we check if [$configuration->getVary() is null](https://github.com/sensiolabs/SensioFrameworkExtraBundle/blob/master/EventListener/HttpCacheListener.php#L122) but it never will be since we [declared the variable to be an array()](https://github.com/sensiolabs/SensioFrameworkExtraBundle/blob/master/Configuration/Cache.php#L57).

### DIFF
--- a/Configuration/Cache.php
+++ b/Configuration/Cache.php
@@ -54,7 +54,7 @@ class Cache extends ConfigurationAnnotation
      *
      * @var array
      */
-    protected $vary = array();
+    protected $vary;
 
     /**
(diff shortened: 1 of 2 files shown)


fabpot:
Thank you @Nyholm.

Nyholm:
Thank you for merging

PR #350, opened 2015-04-02
Remove unused property in the SecurityListener

stof:
``$this->securityContext`` is never accessed anymore. All usages are using the new properties.


fabpot:
Good catch, thanks @stof.

PR #352, opened 2015-05-02
[ParamConverter] made 404 error message configurable

77web:
Hi.
This PR enables customizing 404 error messages in ParamConverter by adding ``message`` configuration.
Usage is shown in doc diff: https://github.com/sensiolabs/SensioFrameworkExtraBundle/compare/sensiolabs:master...77web:paramconverter-configurable-message?expand=1#diff-0b540e20d1b1be77568646ad3de2d31e 

I hope you like it :smile:
Thank you.

### DIFF
--- a/Configuration/ParamConverter.php
+++ b/Configuration/ParamConverter.php
@@ -57,6 +57,13 @@ class ParamConverter extends ConfigurationAnnotation
     protected $converter;
 
     /**
+     * To be used as 404 error message when object is not found
+     *
+     * @var string
+     */
+    protected $message;
+
+    /**
      * Returns the parameter name.
      *
      * @return string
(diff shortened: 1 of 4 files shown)


fabpot:
I'm not I understand why you would want to customize this message. The exception messages are not meant to be displayed to the end user, but only to developers.

77web:
> The exception messages are not meant to be displayed to the end user, but only to developers.

yes, I totally agree with this in most cases. 
but it seemed to be the only way to show different messages to our users by the type of missing resource. (e.g.  an item is missing or a static help page is missing.)

now that I understood using Exception message is not recommended, I'll do it in another way.
maybe I should use app.request something in our 404 page?
thank you anyway.

PR #353, opened 2015-05-02
[Tests] Fixed security.yml path in Extension test when bundled SecurityBundle uses psr-4.

77web:
I happened to find it with test failing in my another PR(#352).
Please confirm and merge. Thank you.


fabpot:
Thank you @77web.

PR #357, opened 2015-05-21
Use PSR-4 instead of PSR-0

jvasseur:
Remove use of deprecated target-dir feature.


jakzal:
:+1:

fabpot:
Thank you @jvasseur.

PR #360, opened 2015-05-26
PSR-7 ParamConverter and view listener.

dunglas:
Replace symfony/symfony#14751

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,5 +20,6 @@ before_install:
   - travis_retry composer self-update
 
 install:
+  - if [[ "$TRAVIS_PHP_VERSION" != "5.3" ]] && [[ "$TRAVIS_PHP_VERSION" != "5.4" ]]; then composer require --no-update symfony/psr-http-message-bridge:dev-wip zendframework/zend-diactoros; fi;
   - if [ "$deps" = "no" ]; then composer --prefer-source install; fi;
   - if [ "$deps" = "low" ]; then composer --prefer-source --prefer-lowest --prefer-stable update; fi;
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,6 @@ before_install:
   - travis_retry composer self-update
 
 install:
-  - if [[ "$TRAVIS_PHP_VERSION" != "5.3" ]] && [[ "$TRAVIS_PHP_VERSION" != "5.4" ]]; then composer require --no-update symfony/psr-http-message-bridge:dev-wip zendframework/zend-diactoros; fi;
+  - if [[ "$TRAVIS_PHP_VERSION" != "5.3" ]] && [[ "$TRAVIS_PHP_VERSION" != "5.4" ]]; then composer require --no-update symfony/psr-http-message-bridge zendframework/zend-diactoros; fi;
   - if [ "$deps" = "no" ]; then composer --prefer-source install; fi;
   - if [ "$deps" = "low" ]; then composer --prefer-source --prefer-lowest --prefer-stable update; fi;
--- a/DependencyInjection/SensioFrameworkExtraExtension.php
+++ b/DependencyInjection/SensioFrameworkExtraExtension.php
@@ -102,6 +102,10 @@ class SensioFrameworkExtraExtension extends Extension
                 $container->getDefinition('sensio_framework_extra.converter.listener')->replaceArgument(1, $config['request']['auto_convert']);
             }
         }
+
+        if ($config['psr_message']['enabled']) {
+            $loader->load('psr7.xml');
+        }
     }
 
     /**
--- a/EventListener/PsrResponseListener.php
+++ b/EventListener/PsrResponseListener.php
@@ -1,0 +1,63 @@
+<?php
+
+/*
+ * This file is part of the Symfony framework.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sensio\Bundle\FrameworkExtraBundle\EventListener;
+
+use Psr\Http\Message\ResponseInterface;
+use Symfony\Bridge\PsrHttpMessage\Factory\HttpFoundationFactory;
+use Symfony\Bridge\PsrHttpMessage\HttpFoundationFactoryInterface;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\HttpKernel\Event\GetResponseForControllerResultEvent;
+use Symfony\Component\HttpKernel\KernelEvents;
+
+/**
+ * Converts PSR-7 Response to HttpFoundation Response using the bridge.
+ *
+ * @author Kévin Dunglas <dunglas@gmail.com>
+ */
+class PsrResponseListener implements EventSubscriberInterface
+{
+    /**
+     * @var HttpFoundationFactoryInterface
+     */
+    private $httpFoundationFactory;
+
+    public function __construct(HttpFoundationFactoryInterface $httpFoundationFactory)
+    {
+        $this->httpFoundationFactory = $httpFoundationFactory;
+    }
+
+    /**
+     * Do the conversion if applicable and update the response of the event.
+     *
+     * @param GetResponseForControllerResultEvent $event
+     */
+    public function onKernelView(GetResponseForControllerResultEvent $event)
+    {
+        $controllerResult = $event->getControllerResult();
+
+        if (!$controllerResult instanceof ResponseInterface) {
+            return;
+        }
+
+        $event->setResponse($this->httpFoundationFactory->createResponse($controllerResult));
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public static function getSubscribedEvents()
+    {
+        return array(
+            KernelEvents::VIEW => 'onKernelView',
+        );
+    }
+}
--- a/Request/ParamConverter/PsrServerRequestParamConverter.php
+++ b/Request/ParamConverter/PsrServerRequestParamConverter.php
@@ -1,0 +1,62 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sensio\Bundle\FrameworkExtraBundle\Request\ParamConverter;
+
+use Sensio\Bundle\FrameworkExtraBundle\Configuration\ParamConverter;
+use Symfony\Bridge\PsrHttpMessage\HttpMessageFactoryInterface;
+use Symfony\Component\HttpFoundation\Request;
+
+/**
+ * Converts HttpFoundation Request to PSR-7 ServerRequest using the bridge.
+ *
+ * @author Kévin Dunglas <dunglas@gmail.com>
+ */
+class PsrServerRequestParamConverter implements ParamConverterInterface
+{
+    /**
+     * @var array
+     */
+    private static $supportedTypes = array(
+        'Psr\Http\Message\ServerRequestInterface' => true,
+        'Psr\Http\Message\RequestInterface' => true,
+        'Psr\Http\Message\HttpMessageInterface' => true,
+    );
+    /**
+     * @var HttpMessageFactoryInterface
+     */
+    private $httpMessageFactory;
+
+    public function __construct(HttpMessageFactoryInterface $httpMessageFactory)
+    {
+        $this->httpMessageFactory = $httpMessageFactory;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function apply(Request $request, ParamConverter $configuration)
+    {
+        $request->attributes->set($configuration->getName(), $this->httpMessageFactory->createRequest($request));
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function supports(ParamConverter $configuration)
+    {
+        if (null === $configuration->getClass()) {
+            return false;
+        }
+
+        return isset(self::$supportedTypes[$configuration->getClass()]);
+    }
+}
--- a/Resources/config/psr7.xml
+++ b/Resources/config/psr7.xml
@@ -5,8 +5,8 @@
            xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
 
     <services>
-        <service id="sensio_framework_extra.psr7.http_message_factory" class="Symfony\Bridge\PsrHttpMessage\Factory\DiactorosFactory" />
-        <service id="sensio_framework_extra.psr7.http_foundation_factory" class="Symfony\Bridge\PsrHttpMessage\Factory\HttpFoundationFactory" />
+        <service id="sensio_framework_extra.psr7.http_message_factory" class="Symfony\Bridge\PsrHttpMessage\Factory\DiactorosFactory" public="false" />
+        <service id="sensio_framework_extra.psr7.http_foundation_factory" class="Symfony\Bridge\PsrHttpMessage\Factory\HttpFoundationFactory" public="false" />
 
         <service id="sensio_framework_extra.psr7.converter.server_request" class="Sensio\Bundle\FrameworkExtraBundle\Request\ParamConverter\PsrServerRequestParamConverter">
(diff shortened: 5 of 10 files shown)


dunglas:
One choice to do:
- make the PSR-7 bridge a hard dependency (user-friendly errors)
- disable the PSR-7 support in the extension if the bridge is not installed

iltar:
What about the first option but the ability to disable it in the configuration? Gives us the option to enable it if we want it.

lsmith77:
FOSRestBundle might also need some tweaks like this.

dunglas:
Finally, adding a hard dependency is not an option because Zend Diactoros requires PHP 5.5.
I've taken the approach of automatically enabling the listener and the converter if the PSR-7 Bridge is installed.

Request/ParamConverter/PsrServerRequestParamConverter.php
```diff
@@ -0,0 +1,63 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function supports(ParamConverter $configuration)
+    {
+        if (null === $configuration->getClass()) {
+            return false;
+        }
+
+        return 'Psr\Http\Message\ServerRequestInterface' === $configuration->getClass();
```
stof: What about supporting the case where someone typehints the parent ``RequestInterface`` too ?

dunglas: This is not supported by the bridge and it will not be easy to do: https://github.com/symfony/psr-http-message-bridge/pull/1/files#diff-7e5b3956b6ae2a5924719b9d4eecb8cdR43

dunglas: Just got what you mean. I'll update the PR to support both.

dunglas:
Added support of `RequestInterface` and  `HttpMessageInterface` injection.

jvasseur:
Injecting on request when type hinting for HttpMessageInterface feels strange since a message can also be a response.

stof:
@jvasseur If the controller only cares about being able to use the HttpMessageInterface methods, passing the Request is perfectly valid. And it is not weird IMO. the input of a controller is always the Request, not the Response (what is weird is the fact to use this typehint though)

dunglas:
Agree with @stof

composer.json
```diff
@@ -19,6 +25,8 @@
         "symfony/security-bundle": "~2.4"
     },
     "suggest": {
+        "symfony/psr-http-message-bridge": "To use the PSR-7 converters",
+        "zendframework/zend-diactoros": "To use the Zend Diactoros factory",
```
lsmith77: not sure if this is necessary since you have this suggest already in the bridge:
https://github.com/dunglas/psr-http-message-bridge/blob/wip/composer.json#L27

DependencyInjection/SensioFrameworkExtraExtension.php
```diff
@@ -32,6 +32,10 @@ public function load(array $configs, ContainerBuilder $container)
         $loader = new XmlFileLoader($container, new FileLocator(__DIR__.'/../Resources/config'));
         $loader->load('services.xml');
 
+        if (class_exists('Symfony\Bridge\PsrHttpMessage\HttpFoundationFactoryInterface')) {
+            $loader->load('psr7.xml');
+        }
```
weaverryan: I think it'd be more natural if we add this as configuration (i.e. maybe I'm using the Psr bridge for middleware, but don't actually want the listeners inside the framework). Something like:

```yml
sensio_framework_extra:
    psr_message:
        enabled: true
```

Or something similar. This isn't quite like the rest of the [config](http://symfony.com/doc/current/bundles/SensioFrameworkExtraBundle/index.html#configuration) in this bundle, but I find that config kind of awkward.

EventListener/PsrResponseListener.php
```diff
@@ -0,0 +1,71 @@
...
+ * @author Kévin Dunglas <dunglas@gmail.com>
+ */
+class PsrResponseListener implements EventSubscriberInterface
+{
+    /**
+     * @var HttpFoundationFactoryInterface
+     */
+    private $httpFoundationFactory;
+
+    public function __construct(HttpFoundationFactoryInterface $httpFoundationFactory = null)
```
weaverryan: If I *do* already have an `HttpFoundationFactoryInterface` registered as a service somewhere, how could I have it injected here? But I *may* be over-thinking a non-real use-case :), and we could decide to add a configuration option (a service id) later without breaking BC.

dunglas: It should be possible to inject it with a compiler pass.

DependencyInjection/SensioFrameworkExtraExtension.php
dunglas: Currently if you add the bridge and doesn't typehint a controller nothing will happen, and if you add the bridge, add the typehint and don't want the param conversion it's kinda strange.

weaverryan: What if I add the bridge and don't type-hint the controller (i.e. I'm using the bridge for another purpose, like middleware). In this case, everything is ok - but it's impossible for me to disable these extra listeners that I'm not using.

stof: All listeners are currently disablable (does this word actually exist?) in this bundle. So it should be the case here too. the default value could depend on the availability of the bridge.

dunglas: ok

Request/ParamConverter/PsrServerRequestParamConverter.php
```diff
@@ -0,0 +1,68 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function supports(ParamConverter $configuration)
+    {
+        if (null === $configuration->getClass()) {
+            return false;
+        }
+
+        return isset(self::$supportedTypes[$configuration->getClass()]);
```
jakzal: Why not making `$supportedTypes` an array of supported types (with no mapping to boolean), and check if a class is `in_array()`?

DavidBadura: because it's faster :wink:

dunglas: @DavidBadura all right.

dunglas:
Added configuration.

.travis.yml
```diff
@@ -20,5 +20,6 @@ before_install:
   - travis_retry composer self-update
 
 install:
+  - if [[ "$TRAVIS_PHP_VERSION" != "5.3" ]] && [[ "$TRAVIS_PHP_VERSION" != "5.4" ]]; then composer require symfony/psr-http-message-bridge:dev-wip zendframework/zend-diactoros; fi;
```
xabbuh: Is there a reason not to have the `--dev` and `--no-update` options here?

EventListener/PsrResponseListener.php
```diff
@@ -0,0 +1,71 @@
...
+    {
+        if (null === $httpFoundationFactory) {
+            if (!class_exists('Symfony\Bridge\PsrHttpMessage\Factory\HttpFoundationFactory')) {
+                throw new \RuntimeException('Install the PSR-7 bridge to return a Psr\Http\Message\ResponseInterface instance in the controller.');
+            }
```
xabbuh: If I don't miss anything, this can never happen (the method is type-hinted with `HttpFoundationFactoryInterface` which is in the same package as `HttpFoundationFactory`).

.travis.yml
dunglas: `--dev` is implicit isn't it?
`--no-update` is indeed necessary.

EventListener/PsrResponseListener.php
dunglas: No error will be thrown if the class in the typehint is not found and null (or no value) is passed. This is just a hint.

This case can happen if the support is enabled through configuration, PSR-7 interfaces package is installed, but the bridge is not.

Request/ParamConverter/PsrServerRequestParamConverter.php
```diff
@@ -0,0 +1,68 @@
...
+     * @var HttpMessageFactoryInterface
+     */
+    private $httpMessageFactory;
+
+    public function __construct(HttpMessageFactoryInterface $httpMessageFactory = null)
+    {
+        if (null === $httpMessageFactory) {
+            if (!class_exists('Symfony\Bridge\PsrHttpMessage\Factory\DiactorosFactory')) {
+                throw new \RuntimeException('Install the PSR-7 bridge to inject a Psr\Http\Message\ServerRequestInterface instance in the controller.');
+            }
```
xabbuh: This can never happen for the same reasons as above.

dunglas: It can (see previous comment).

Request/ParamConverter/PsrServerRequestParamConverter.php
```diff
@@ -0,0 +1,68 @@
...
+ *
+ * @author Kévin Dunglas <dunglas@gmail.com>
+ */
+class PsrServerRequestParamConverter implements ParamConverterInterface
+{
+    private static $supportedTypes = [
+        'Psr\Http\Message\ServerRequestInterface' => true,
+        'Psr\Http\Message\RequestInterface' => true,
+        'Psr\Http\Message\HttpMessageInterface' => true,
+    ];
```
xabbuh: We should use the `array()` syntax here. Even if you won't be able to use this annotation with PHP 5.3 or PHP 5.4, you should get a more sensible error message (the exception thrown by the `DiactorosFactory` from the bridge) instead of a parse error.

dunglas: Goof catch!

The converter will work with PHP 5.3 as long a you use (create for now) a PSR-7 implementation compliant with it. The only package requiring PHP 5.5+ is Diactoros and it's a soft dependency.

.travis.yml
xabbuh: `--dev` is only implicit with the `install` and `update` commands. By default, the dependency would be added to the `require` section, but not to the `require-dev` section. This shouldn't make any difference here I guess, but technically it's a dev dependency.

EventListener/PsrResponseListener.php
xabbuh: Hm indeed, you're right. But shouldn't we provide a more meaningful message then if the user enables PSR-7 support while the bridge is not installed (in the `Configuration` class)?

xabbuh: Anyway, we should leave the message here too in case someone uses the class on their own.

What I am more concerned about is the idea to make the argument optional. Is that really a good idea? Why not injecting an `HttpFoundationFactory` instance by default? People wanting to use a different implementation will have to inject it on their own anyway.

.travis.yml
dunglas: It's not a dev dependency if you use the PSR-7 converter / bridge, and in that context we use the bridge.

EventListener/PsrResponseListener.php
dunglas: That code directly come from the PR on Symfony where we couldn't rely on a the DependencyInjection component.
But here we can, so let's go your way, it's cleaner.

.travis.yml
xabbuh: But if we put it in `composer.json`, we would have used `require-dev` as it is no mandatory dependency.

Request/ParamConverter/PsrServerRequestParamConverter.php
jakzal: It might be faster, but it's misleading. The array stores key value pairs, but values are never used.

apfelbox: Not sure whether performance counts as an argument for an array with 3 entries...

dunglas: A call potentially executed several times (when sub request and `render()` are used) on every requests.

stof: and it is called for each argument of the controller

dunglas:
* [x] Use the DI component to inject factories.
* [x] Minor fixes.

Request/ParamConverter/PsrServerRequestParamConverter.php
jakzal: I'm still :-1: as I prefer readability over speculative performance benefit. Few calls during a request is not going to  make any difference ([isset](http://3v4l.org/SR0PP/perf#tabs) vs [in_array](http://3v4l.org/DkSUH/perf#tabs) - only hhvm exhibits visible difference in this silly test).

dunglas:
Added some doc.

Resources/config/psr7.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" ?>
+
+<container xmlns="http://symfony.com/schema/dic/services"
+           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+           xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
+
+    <services>
+        <service id="sensio_framework_extra.psr7.http_message_factory" class="Symfony\Bridge\PsrHttpMessage\Factory\DiactorosFactory" />
+        <service id="sensio_framework_extra.psr7.http_foundation_factory" class="Symfony\Bridge\PsrHttpMessage\Factory\HttpFoundationFactory" />
```
xabbuh: both definitions should be private

Resources/doc/index.rst
```diff
@@ -57,11 +57,13 @@ The default configuration is as follow:
...
-            request:  { converters: true, auto_convert: true }
-            view:     { annotations: true }
-            cache:    { annotations: true }
-            security: { annotations: true }
+            router:      { annotations: true }
+            request:     { converters: true, auto_convert: true }
+            view:        { annotations: true }
+            cache:       { annotations: true }
+            security:    { annotations: true }
+            psr_message: { enabled: false } # Default to true if the PSR-7 bridge is installed
```
xabbuh: defaults

Resources/doc/index.rst
```diff
@@ -72,17 +74,19 @@ The default configuration is as follow:
             <view annotations="true" />
             <cache annotations="true" />
             <security annotations="true" />
+            <psr-message enabled="true" /> <!-- Default to true if the PSR-7 bridge is installed -->
```
xabbuh: defaults

Resources/doc/index.rst
xabbuh: should imho be set to `true` to be consistent with the XML and PHP example

Resources/doc/index.rst
```diff
@@ -184,4 +188,45 @@ example:
 
 see :ref:`Annotated Routes Activation<frameworkextra-annotations-routing-activation>` for more details.
 
+PSR-7 support
+-------------
+
+SensioFrameworkExtraBundle provides support for HTTP messages interfaces defined
+in `PSR-7`_. It allows to inject instances of :class:`Psr\\Http\\Message\\ServerRequestInterface`
+and to return instances of :class:`Psr\\Http\\Message\\ResponseInterface` in controllers.
```
xabbuh: Using the `class` for external class will not work as their API docs are not rendered on symfony.com.

dunglas:
Fixed issues raised by @xabbuh

Tests/EventListener/PsrResponseListenerTest.php
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    public function testConvertsControllerResult()
+    {
+        $event = $this->createEventMock(new Response());
+        $event->expects($this->once())->method('setResponse')->with($this->isInstanceOf('Symfony\Component\HttpFoundation\Response'));
+        $this->listener->onKernelView($event);
+    }
+
+    public function testDoesNtConvertControllerResult()
```
jakzal: *DoesNot*

jakzal:
Great job :+1:

Request/ParamConverter/PsrServerRequestParamConverter.php
jakzal: Since two other core contributors defend the current way, I won't be blocking this PR.

dunglas: As variable is private, we can style update our CS once merged to allow or disallow that kind of tricks and change it here if necessary.

dunglas:
@fabpot should be ok now

fabpot:
Thank you @dunglas.

Tobion:
This now requires Diactoros even if you only want the PSR-7 -> HttpFoundation transformation: #406

PR #363, opened 2015-06-03
Rename HttpMessageInterface to MessageInterface (PSR-7)

barryvdh:
HttpMessageInterface doesn't exist, see https://github.com/php-fig/http-message/blob/master/src/MessageInterface.php


xabbuh:
:+1:

fabpot:
Thank you @barryvdh.

Issue #364, opened 2015-07-07
@Route(service="...") triggers deprecation notices as of 2.7 


iltar:
I'm using Controller as a Service. Therefore I put the following annotations on my controllers: `@Route(service="app.controller.test")`. This is all fine and such, but the `Symfony\Component\Routing\Loader\AnnotationClassLoader` cannot deal with this properly due to BC reasons. The following snippet should explain more:

https://github.com/symfony/Routing/blob/2.7/Loader/AnnotationClassLoader.php#L223-L228
```php
// for BC reasons
if (null !== $annot->getPath()) {
    $globals['path'] = $annot->getPath();
} elseif (null !== $annot->getPattern()) {
    $globals['path'] = $annot->getPattern();
}
```
With the annotation provided, this will trigger the following deprecation warnings (1x per controller in the app):
```
DEPRECATED - The Symfony\Component\Routing\Annotation\Route::getPattern method is deprecated since version 2.2 and will be removed in 3.0. Use the getPath() method instead and use the "path" option instead of the "pattern" option in the route definition. DEPRECATED - The Symfony\Component\Routing\Annotation\Route::getPattern method is deprecated since version 2.2 and will be removed in 3.0. Use the getPath() method instead and use the "path" option instead of the "pattern" option in the route definition.
```

A quickfix would be to change all the annotations to `@Route("", service="app.controller.test")`. Now you can understand that this is not really the most wanted solution...

A few proposals:
 - [Change how Controller as a Service is registered.](https://github.com/symfony/symfony/issues/12915) 
 - Modify the setter to set `$path` to an empty string upon setting `$service` if it's null.
 - Setting the default value of `$path` to an empty string

The last two solutions are probably the easiest fix but I'm not sure if this fixes design issue of re-using the `Route` annotations for something it shouldn't do.

weaverryan:
@iltar I'd go for number (2) and make a pull request for this. I think that's the only solution, or don't fix this (and have people work around the deprecation warning with the quickfix you listed).

PR #366, opened 2015-07-16
Fix a BC notice when route path is empty in sf 2.7

iltar:
This PR fixes Issue #364 via the suggestion of @weaverryan. When `@Route(service="app.test")` is called, Symfony 2.7 will trigger a deprecation notice because it has no `path`, which in turns calls a deprecated getter after that. Now `setService` will call `setPath` if it wasn't set yet and fill it with an empty string to avoid the getter from being called.

### DIFF
--- a/Configuration/Route.php
+++ b/Configuration/Route.php
@@ -23,6 +23,10 @@ class Route extends BaseRoute
 
     public function setService($service)
     {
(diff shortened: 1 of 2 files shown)


fabpot:
Thank you @iltar.

PR #368, opened 2015-07-31
[DX] Added safeguard to avoid silently overriding configuration that disallows array

hshn:
Hi,

When annotating multiple configuration that disallow array (such as `@Security`) to one action, previous ones will be ignored silently. It looked like not working to me when I experienced first time.

I think it will be better to have an ability to notice that it is an incorrect configurations.
thanks

### DIFF
--- a/EventListener/ControllerListener.php
+++ b/EventListener/ControllerListener.php
@@ -93,8 +93,10 @@ class ControllerListener implements EventSubscriberInterface
             if ($configuration instanceof ConfigurationInterface) {
                 if ($configuration->allowArray()) {
                     $configurations['_'.$configuration->getAliasName()][] = $configuration;
-                } else {
+                } elseif (!isset($configurations['_'.$configuration->getAliasName()])) {
                     $configurations['_'.$configuration->getAliasName()] = $configuration;
+                } else {
+                    throw new \LogicException(sprintf('Configure "%s" multiple is not allowed', $configuration->getAliasName()));
                 }
             }
         }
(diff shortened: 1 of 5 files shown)


EventListener/ControllerListener.php
```diff
@@ -93,8 +93,10 @@ protected function getConfigurations(array $annotations)
             if ($configuration instanceof ConfigurationInterface) {
                 if ($configuration->allowArray()) {
                     $configurations['_'.$configuration->getAliasName()][] = $configuration;
-                } else {
+                } elseif (!isset($configurations['_'.$configuration->getAliasName()])) {
                     $configurations['_'.$configuration->getAliasName()] = $configuration;
+                } else {
+                    throw new \LogicException(sprintf('Configure "%s" multiple is not allowed', $configuration->getAliasName()));
```
fabpot: I don't understand the error message.

fabpot: `Multiple "%s" annotations are not allowed`?

fabpot:
Thank you @hshn.

hshn:
Thank you for merging this. And sorry for my bad english. I will try to improve it.

RowanReid:
I am using this bundle and am now experiencing this new Exception.  Thus this updated is doing it's job of notifying me that my configurations are incorrect!

That said, given that multiple @Security annotations are not allowed, how does one check multiple roles, or check the is_granted() status on multiple action parameters?  It is not clear from the documentation how this is done.

RowanReid:
Ok I have answered this myself:

```php
/**
 * @Security("is_granted('POST_SHOW', post) && as_role('ROLE_ADMIN')")
 */
{
}
```
This would be a nice addition to the official documentation: http://symfony.com/doc/current/bundles/SensioFrameworkExtraBundle/annotations/security.html

fabpot:
@RowanReid Can you propose an addition to the docs as a pull request?

RowanReid:
Done https://github.com/sensiolabs/SensioFrameworkExtraBundle/pull/370

PR #370, opened 2015-08-03
Added an additional usage example

RowanReid:
Added an additional example to illustrate how to make multiple function calls in the expression.


fabpot:
Thank you @RowanReid.

PR #371, opened 2015-08-31
Add new ParamConverter which creates entity objects based on Request …

marius-alex:
I noticed that the existing bundle works only for retrieving objects from Doctrine but does not handle the creation of new entities which is usefull for POST/PUT actions.

The converter has the following features:
- persist or not the object based on an optional configuration
- has excludes and mapping like his counterpart
- supports Symfony Validators


fabpot:
See #783

PR #374, opened 2015-09-22
Add an option to throw NotFoundHttpException

iamluc:
…instead of AccessDeniedException.

For security reason, it is sometime better to return status code 404 when a user doesn't have access to a resource to prevent leaking private informations.

This PR adds a new option on the `@Security` annotation

```
    /**
     * @Security("is_granted('POST_SHOW', post)", notFound=true)
     */
    {
    }
```

### DIFF
--- a/EventListener/SecurityListener.php
+++ b/EventListener/SecurityListener.php
@@ -13,6 +13,7 @@ namespace Sensio\Bundle\FrameworkExtraBundle\EventListener;
 
 use Sensio\Bundle\FrameworkExtraBundle\Security\ExpressionLanguage;
 use Symfony\Component\HttpKernel\Event\FilterControllerEvent;
+use Symfony\Component\HttpKernel\Exception\HttpException;
 use Symfony\Component\HttpKernel\KernelEvents;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
@@ -65,7 +66,12 @@ class SecurityListener implements EventSubscriberInterface
         }
 
         if (!$this->language->evaluate($configuration->getExpression(), $this->getVariables($request))) {
-            throw new AccessDeniedException(sprintf('Expression "%s" denied access.', $configuration->getExpression()));
+            if ($statusCode = $configuration->getStatusCode()) {
+                throw new HttpException($statusCode, $configuration->getMessage());
+            }
+
+            $message = $configuration->getMessage() ?: sprintf('Expression "%s" denied access.', $configuration->getExpression());
+            throw new AccessDeniedException($message);
         }
     }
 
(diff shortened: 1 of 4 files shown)


hhamon:
Can't we have a more generic `statusCode` option instead which is set to `403` by default? So that we can change it and raise a `400` or `404`. The system will basically throw a regular `HttpException`.

iamluc:
Hi @hhamon 
Thanks for your quick reply.

I just updated the PR to use `statusCode`. I also added a `message` option to specify the message of the exception.

hhamon:
Nice :)

EventListener/SecurityListener.php
```diff
@@ -65,7 +66,12 @@ public function onKernelController(FilterControllerEvent $event)
         }
 
         if (!$this->language->evaluate($configuration->getExpression(), $this->getVariables($request))) {
-            throw new AccessDeniedException(sprintf('Expression "%s" denied access.', $configuration->getExpression()));
+            if ($statusCode = $configuration->getStatusCode()) {
+                throw new HttpException($statusCode, $configuration->getMessage());
```
xabbuh: We need a default message in case someone only modified the status code.

iamluc: Do you have a suggestion ?
I'm not sure `Expression "%s" denied access.` fits for all case

fabpot: I would just use `Access denied` as a default message.

EventListener/SecurityListener.php
```diff
@@ -65,7 +66,12 @@ public function onKernelController(FilterControllerEvent $event)
...
+            }
+
+            $message = $configuration->getMessage() ?: sprintf('Expression "%s" denied access.', $configuration->getExpression());
```
fabpot: could be inlined.

fabpot: I would never use `Expression "%s" denied access.` as a message, but the more generic `Access denied` one.

Resources/doc/annotations/security.rst
```diff
@@ -54,6 +54,26 @@ Here is another example, making use of multiple functions in the expression::
     {
     }
 
+You can throw an exception ``Symfony\Component\HttpKernel\Exception\HttpException``
```
fabpot: You can throw an HttpException exception instead of

Resources/doc/annotations/security.rst
```diff
@@ -54,6 +54,26 @@ Here is another example, making use of multiple functions in the expression::
...
+using the ``statusCode`` option::
+
+    /**
+     * @Security("is_granted('POST_SHOW', post)", statusCode=404)
+     */
+    public function showAction(Post $post)
+    {
+    }
+
+Option ``message`` will allow you to specify the message of the exception::
```
fabpot: allows you to

fabpot:
I cannot push on this PR, so I'm going to close it and submit a new one with some tweaks.

fabpot:
see #510

PR #376, opened 2015-10-08
Adds not_found_message configuration to Doctrine ParamConverter

makarova:
Adds possibility to customize not found exception message in doctrine param converter by using not_found_message configuration property

### DIFF
--- a/Request/ParamConverter/DoctrineParamConverter.php
+++ b/Request/ParamConverter/DoctrineParamConverter.php
@@ -63,7 +63,13 @@ class DoctrineParamConverter implements ParamConverterInterface
         }
 
         if (null === $object && false === $configuration->isOptional()) {
-            throw new NotFoundHttpException(sprintf('%s object not found.', $class));
+
+            $message = sprintf('%s object not found.', $class);
+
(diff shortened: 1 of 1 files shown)


apfelbox:
See #196

makarova:
@apfelbox I saw that issue but still think that it is very convenient to have possibility to customize error message, especially if it is so simple to implement

ghanbari:
I use multiply `@ParamConvertor` on my method and i want show a appropriate message, but i can not detect in `404.json.twig`, which object wasn't found,

also i use **fos rest bundle** that permit to me show exception message to user, but i no want show class name to final user, (that is a good ability, if and if i can customize exception message)

ghanbari:
I use multiply `@ParamConvertor` on my method and i want show a appropriate message, but i can not detect in `404.json.twig`, which object wasn't found,

also i use **fos rest bundle** that permit to me show exception message to user, but i no want show class name to final user, (that is a good ability, if and if i can customize exception message)

fabpot:
Closing as exception messages are not meant to be displayed to the user.

PR #378, opened 2015-10-28
remove build job for nightly PHP build

xabbuh:
The `nightly` PHP version on Travis CI is no longer PHP 7.0 but the nightly build of PHP 7.1.

Furthermore, builds on PHP 7 should no longer be allowed to fail as PHP 7 is almost stable and we should rather report bugs to PHP itself instead of silently ignore such errors.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,15 +11,13 @@ php:
   - 5.4
   - 5.5
   - 5.6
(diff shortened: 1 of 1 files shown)


fabpot:
Thank you @xabbuh.

PR #379, opened 2015-10-31
Custom 'not found message' for 'ParamConvertor' when object is not found

ghanbari:
Custom 'not found message' for 'ParamConvertor' when object is not found


ghanbari:
close on [duplicate](https://github.com/sensiolabs/SensioFrameworkExtraBundle/pull/376)

Issue #312, opened 2014-06-24
@Method({"POST"}) throws MethodNotAllowedException


komex:
A few days ago this code works fine, but now it always thows `MethodNotAllowedException` on `POST` HTTP request.

``` php
use Sensio\Bundle\FrameworkExtraBundle\Configuration\Method;

/**
 * @Permissions(roles={"ROLE_MODERATOR"})
 * @Route("/hcs-catalog/element/delete/{element}", name="cms_hcs_catalog_element_delete")
 * @Method({"POST"})
 */
public function deleteElementAction(Request $request, HCSElements $element)
{
    if (!$request->isXmlHttpRequest()) {
        throw new BadRequestHttpException('Allowed only AJAX requests.');
    }
    $this->getModel()->deleteElement($element);

    return new Response();
}
```

PR #380, opened 2015-11-04
[Routing] fix to allow verbs http annotation @Method("GET|POST")

sergiors:
Inside $route->setMethods() is called implode()
I think this issue solve #312

### DIFF
--- a/Routing/AnnotatedRouteControllerLoader.php
+++ b/Routing/AnnotatedRouteControllerLoader.php
@@ -50,7 +50,7 @@ class AnnotatedRouteControllerLoader extends AnnotationClassLoader
         // requirements (@Method)
         foreach ($this->reader->getMethodAnnotations($method) as $configuration) {
             if ($configuration instanceof Method) {
-                $route->setMethods($configuration->getMethods());
+                $route->setMethods(explode('|', implode('|', $configuration->getMethods())));
             } elseif ($configuration instanceof FrameworkExtraBundleRoute && $configuration->getService()) {
                 throw new \LogicException('The service option can only be specified at class level.');
             }
--- a/Routing/AnnotatedRouteControllerLoader.php
+++ b/Routing/AnnotatedRouteControllerLoader.php
@@ -50,7 +50,8 @@ class AnnotatedRouteControllerLoader extends AnnotationClassLoader
         // requirements (@Method)
         foreach ($this->reader->getMethodAnnotations($method) as $configuration) {
             if ($configuration instanceof Method) {
-                $route->setMethods(explode('|', implode('|', $configuration->getMethods())));
+                $methods = implode('|', $configuration->getMethods());
+                $route->setMethods(explode('|', $methods));
             } elseif ($configuration instanceof FrameworkExtraBundleRoute && $configuration->getService()) {
                 throw new \LogicException('The service option can only be specified at class level.');
             }
--- a/Routing/AnnotatedRouteControllerLoader.php
+++ b/Routing/AnnotatedRouteControllerLoader.php
@@ -50,7 +50,7 @@ class AnnotatedRouteControllerLoader extends AnnotationClassLoader
         // requirements (@Method)
         foreach ($this->reader->getMethodAnnotations($method) as $configuration) {
             if ($configuration instanceof Method) {
-                $route->setMethods(implode('|', $configuration->getMethods()));
+                $route->setMethods($configuration->getMethods());
             } elseif ($configuration instanceof FrameworkExtraBundleRoute && $configuration->getService()) {
                 throw new \LogicException('The service option can only be specified at class level.');
             }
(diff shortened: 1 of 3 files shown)


Routing/AnnotatedRouteControllerLoader.php
```diff
@@ -50,7 +50,7 @@ protected function configureRoute(Route $route, \ReflectionClass $class, \Reflec
         // requirements (@Method)
         foreach ($this->reader->getMethodAnnotations($method) as $configuration) {
             if ($configuration instanceof Method) {
-                $route->setMethods(implode('|', $configuration->getMethods()));
+                $route->setMethods(explode('|', implode('|', $configuration->getMethods())));
```
xabbuh: What is the point in calling `implode()` at all if we just call `explode()` with its result?

sergiors: it's ugly, i know.. hehhe
OK, `$configuration->getMethods()` may return 
```
[
     "GET",
     "PUT|POST",
]
```

```
implode('|', $methods); \\ => "GET|PUT|POST"
``` 
as `$route-> setMethods()`works with array, I call explode to normalize

sergiors: `explode` needs a string and `$configuration->getMethods()` return just array

sergiors: @xabbuh ping

xabbuh: Ah I see, makes sense then.

xabbuh: Or can we maybe just change the configuration to already return a flat list of methods?

sergiors: Normalize the array in https://github.com/sensiolabs/SensioFrameworkExtraBundle/blob/master/Configuration/Method.php#L46

?

xabbuh: Looks like it's not needed there. So when could `$configuration->getMethods()` actually return what you posted above?

sergiors: Yes, `$configuration->getMethods()` could return
```
[
     "PUT|POST",
]
```

https://github.com/sensiolabs/SensioFrameworkExtraBundle/pull/380/files#diff-f23783624ea580a4daab1fd95f6a31c0R22

xabbuh: In that line it already is a sequence and is https://github.com/sensiolabs/SensioFrameworkExtraBundle/pull/380/files#diff-f23783624ea580a4daab1fd95f6a31c0R30 really a supported use case?

xabbuh: Under which circumstances?

sergiors: https://github.com/sensiolabs/SensioFrameworkExtraBundle/pull/380/files#diff-f23783624ea580a4daab1fd95f6a31c0R22

xabbuh: Sorry, but I cannot reproduce your observations. That example gives me the following output:

```
Array
(
    [0] => GET
    [1] => POST
)
```

Which version of the bundle did you use (I tested it with the latest version, 3.0.11)?

sergiors: @xabbuh the last version..
I cloned the bundle and add https://github.com/sergiors/SensioFrameworkExtraBundleRouteAnnotation/commit/adc01bb0723805d67e9a2730ddafee8f97fa6f5c

run: `composer install` and `phpunit tests/Routing/AnnotatedRouteControllerLoaderTest.php`

the result:
```
PHPUnit 5.0.0 by Sebastian Bergmann and contributors.

..F                                                                 3 / 3 (100%)

Time: 144 ms, Memory: 13.50Mb

There was 1 failure:

1) Sensio\Bundle\FrameworkExtraBundle\Tests\Routing\AnnotatedRouteControllerLoaderTest::testLoad
Failed asserting that two arrays are equal.
--- Expected
+++ Actual
@@ @@
 Array (
-    0 => 'GET'
-    1 => 'POST'
+    0 => 'GET|POST'
 )
```

apfelbox:
@sergiors you read the error message from PhpUnit wrong.

It expected to find `["GET", "POST"]` but it found `["GET|POST"]` (not an array with 3 items. Please look at `-` and `+` at the beginning of the line.

sergiors:
@apfelbox do you run https://github.com/sensiolabs/SensioFrameworkExtraBundle/pull/380#discussion_r45276564 ?

apfelbox:
@sergiors my comment was only pointed at your comment

> ```
> [
>     'GET',
>     'POST',
>     'GET|POST'
> ]
> ```

which is a wrong interpretation of the output from phpunit.

I ran your example and `@Method({"GET", "POST"})` returns `array('GET|POST')`, like it is expected to do. Not sure what you are trying to fix.

As far as I can see, the annotation always only returns an array with exactly one element (`"GET"` or `"GET|POST"`). Do you have a case where it does something different?

PR #381, opened 2015-11-04
Fix ControllerListener annotations retrieval for invokable controllers

ogizanagi:
Fixes https://github.com/symfony/symfony/issues/16451

However, I think that for such controllers, having the annotation at the class level makes much sense.

### DIFF
--- a/EventListener/ControllerListener.php
+++ b/EventListener/ControllerListener.php
@@ -50,15 +50,16 @@ class ControllerListener implements EventSubscriberInterface
      */
     public function onKernelController(FilterControllerEvent $event)
     {
-        if (!is_array($controller = $event->getController())) {
+        if (!is_array($controller = $event->getController()) && !is_object($controller)) {
             return;
         }
 
-        $className = class_exists('Doctrine\Common\Util\ClassUtils') ? ClassUtils::getClass($controller[0]) : get_class($controller[0]);
-        $object = new \ReflectionClass($className);
-        $method = $object->getMethod($controller[1]);
+        list($object, $methodName) = is_array($controller) ? $controller : array($controller, '__invoke');
+        $className = class_exists('Doctrine\Common\Util\ClassUtils') ? ClassUtils::getClass($object) : get_class($object);
+        $class = new \ReflectionClass($className);
+        $method = $class->getMethod($methodName);
 
-        $classConfigurations = $this->getConfigurations($this->reader->getClassAnnotations($object));
+        $classConfigurations = $this->getConfigurations($this->reader->getClassAnnotations($class));
         $methodConfigurations = $this->getConfigurations($this->reader->getMethodAnnotations($method));
 
         $configurations = array();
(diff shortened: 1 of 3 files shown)


EventListener/ControllerListener.php
```diff
@@ -50,13 +50,18 @@ public function __construct(Reader $reader)
...
 
-        $className = class_exists('Doctrine\Common\Util\ClassUtils') ? ClassUtils::getClass($controller[0]) : get_class($controller[0]);
-        $object = new \ReflectionClass($className);
-        $method = $object->getMethod($controller[1]);
+        if (is_array($controller)) {
+            $className = class_exists('Doctrine\Common\Util\ClassUtils') ? ClassUtils::getClass($controller[0]) : get_class($controller[0]);
+            $object = new \ReflectionClass($className);
+            $method = $object->getMethod($controller[1]);
+        } else {
+            $object = new \ReflectionClass(get_class($controller));
```
Taluu: `new ReflectionObject($controller)` ?

EventListener/ControllerListener.php
```diff
@@ -50,13 +50,18 @@ public function __construct(Reader $reader)
      */
     public function onKernelController(FilterControllerEvent $event)
     {
-        if (!is_array($controller = $event->getController())) {
+        if (!is_array($controller = $event->getController()) && !is_object($controller)) {
```
stof: this won't deal properly with closures, which are objects (even though they are unlikely to be used in a fullstack project, and the bundle is unlikely to be used elsewhere)

EventListener/ControllerListener.php
stof: getting the class name should use the ClassUtils too, to account for proxy classes (JMSDiExtraBundle generates such proxies for instance).

@Taluu no, this would get the wrong method for proxy objects (the current code is also broken)

EventListener/ControllerListener.php
ogizanagi: closures are objects and defines an `__invoke` method. Of course there won't be any annotations, but is it worth it to avoid passing into this listener for such cases ? As you said, it's unlikely to happen, and it is actually "properly" handled (won't break).

Or maybe I missed the point ?

EventListener/ControllerListener.php
```diff
@@ -50,13 +50,14 @@ public function __construct(Reader $reader)
...
             return;
         }
 
-        $className = class_exists('Doctrine\Common\Util\ClassUtils') ? ClassUtils::getClass($controller[0]) : get_class($controller[0]);
+        list($class, $methodName) = is_array($controller) ? $controller : array($controller, '__invoke');
```
stof: ``$class`` does not always store a class here. Most of the time, it is an object

ogizanagi: Ok. I reversed `object` and `class`, respectively for the controller object and the reflection class. Let me know if you think there is even better. Thanks.

EventListener/ControllerListener.php
iltar: In case of anonymous functions (I think that's what @stof is aiming at), they cannot have Annotations and should therefore be skipped in this listener. I think that the only format that should be supported is `[controller, action]` where `action` can be `*Action` or `__invoke`. The first is not mandatory (I think) but is a requirement for auto-guessing template names.

Ener-Getick:
Also fixed by https://github.com/sensiolabs/SensioFrameworkExtraBundle/pull/398/files

PR #387, opened 2015-12-04
Fixed documentation example of @Cache ETag field

o:
Hello!

Currently i'm working with Symfony 2.7 and tried to use @Cache example from documentation. Current example doesn't work. 

`ETag` field on annotation used with concat and Expression language expects string for concat operation, but example written for using with `\DateTime` object. (`setLastModifed` already expects `\DateTime` object)

With current example i'm getting `Catchable Fatal Error: Object of class DateTime could not be converted to string`. So i changed example with working version.


jakzal:
:+1:

fabpot:
Thank you @o.

PR #389, opened 2015-12-12
Remove deprecated use of SecurityContextInterface.

AdrienLucas:
This is similar to https://github.com/sensiolabs/SensioFrameworkExtraBundle/pull/388 but it goes further by removing all references to the old SecurityContextInterface. I'm not sure it's not breaking BC, so I keep the two PR open for now.


AdrienLucas:
I added the same commit as in https://github.com/sensiolabs/SensioFrameworkExtraBundle/pull/388 to skip the test if it executes on an old symfony stack (with no TokenStorageInterface). Please tell me if it's the way to go.

AdrienLucas:
This was a misunderstanding : I didn't noticed that this bundle should be working with a 2.3 stack.
Closing the PR.

Issue #385, opened 2015-11-20
@Template does not work when there are no variables


mediamonks-robert:
When using @Template it is apparently required to set any variable or otherwise it will silently ignore and throw an exception about no response being returned.

```php
class DefaultController extends Controller
{
    /**
     * @Template
     */
    public function indexAction()
    {
    }
}
```

Expected result: template MyBundle:Default:index.html.twig will be rendered
Actual result: an exception is thrown: "The controller must return a response (null given). Did you forget to add a return statement somewhere in your controller?"

The code responsible:

https://github.com/sensiolabs/SensioFrameworkExtraBundle/blob/master/EventListener/TemplateListener.php#L98

It is not clear from documentation that having a variable is actually a requirement, I would expect it to work without too. My preference is that a template is being rendered anyway without this current checks on template vars, if this is not possible it would be good to put it in documentation.

iltar:
This case is only triggered if you return null _and_ have no arguments. I can make a fix for this as my intention was to work on this either way.

jakzal:
You need to at least return an empty array.

iltar:
I'm working on a PR and take this into consideration (I need to refactor quite a bit in there).

PR #391, opened 2015-12-18
Reduce reflection calls from 3 to 2 in most cases

iltar:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #385
| License       | MIT
| Doc PR        | ~

This PR aims to fix the issue where reflection was called to calculate the method signature (arguments) in the controller event. Those parameters were only ever used [if the controller returns "null" as response](https://github.com/sensiolabs/SensioFrameworkExtraBundle/blob/v3.0.11/EventListener/TemplateListener.php#L95) in the view event. Instead of doing most of it in the controller event, this resolving is now done in the view event instead and prevents a reflection call in most cases. [Previously it was calculated here.](https://github.com/sensiolabs/SensioFrameworkExtraBundle/blob/v3.0.11/EventListener/TemplateListener.php#L72)

Additionally I've fixed the issue reported here in #385. If you would pass no arguments and return null, it would never render and fail instead. Now it just renders the template without any additional parameters.

Because this class was untested, I've taken the liberty to add a bunch of functional tests that test most of the code (happy flow). I have not yet had time to properly test the unhappy flow and can be done if it's a requirement of the PR. Those tests should help with the annotations in general as also the `@Route` is tested here.

~~#388 is blocking the CI for php 5.5+~~

### DIFF
--- a/Configuration/Template.php
+++ b/Configuration/Template.php
@@ -24,7 +24,7 @@ class Template extends ConfigurationAnnotation
     /**
      * The template reference.
      *
-     * @var TemplateReference
+     * @var TemplateReference|string
      */
     protected $template;
 
@@ -48,6 +48,13 @@ class Template extends ConfigurationAnnotation
      * @var bool
      */
     protected $streamable = false;
+
+    /**
+     * The controller (+action) this annotation is set to.
+     *
+     * @var array
+     */
+    private $owner;
 
     /**
      * Returns the array of templates variables.
@@ -157,5 +164,23 @@ class Template extends ConfigurationAnnotation
     public function allowArray()
     {
         return false;
+    }
+
+    /**
+     * @param array $owner
+     */
+    public function setOwner(array $owner)
+    {
+        $this->owner = $owner;
+    }
+
+    /**
+     * The controller (+action) this annotation is attached to
+     *
+     * @return array
+     */
+    public function getOwner()
+    {
+        return $this->owner;
     }
 }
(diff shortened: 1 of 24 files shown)


iltar:
Had some issues with the deps=low. It turns out that some of the dependencies were not configured but indirectly gotten from the framework bundle.
 - crawler is 2.0 by default
 - dependency injection 2.2
 - twig bundle was missing altogether

Minimal requirements are 2.3 for the first two and 1.11 for twig/twig.

iltar:
ping @fabpot Is there anything missing in this PR to be merged? 

I hate to ping you directly but it seems like this repository isn't getting the same love as symfony does ._.

fabpot:
Thank you @iltar.

fabpot:
Sorry for the delay. It's fine to ping me, there are so many PRs happening on all my projects that sometimes a small ping for important PRs is useful.

Configuration/Template.php
```diff
@@ -158,4 +165,22 @@ public function allowArray()
     {
         return false;
     }
+
+    /**
+     * @param array $owner
+     */
+    public function setOwner(array $owner)
```
stof: technically, this is not always an array (you even removed the condition ensuring that non-array callables are not handled anymore by the listener).

iltar: As far as I can tell, this is never triggered if it's not an array: https://github.com/sensiolabs/SensioFrameworkExtraBundle/blob/master/EventListener/ControllerListener.php#L51-L55

PR #394, opened 2016-01-14
Allow multiple @Security

lyrixx:
The current behavior is not cool because it allows user to use only one
`@Security` tag per action OR per class. BUT Nothing prevents user to put
a tag on the class AND another one on the action. And it's even worst
because only the tag on the action will be used.

Because almost all annotations allow multiple instances (like `@Route`,
`@Template`, `@Method`) I think it's really convenient to allow multiple
`@Security`, and more over to fix a bug.

I already see theses PR;
* https://github.com/sensiolabs/SensioFrameworkExtraBundle/pull/368
* https://github.com/sensiolabs/SensioFrameworkExtraBundle/pull/370

But they are not perfect.

### DIFF
--- a/Configuration/Security.php
+++ b/Configuration/Security.php
@@ -43,6 +43,6 @@ class Security extends ConfigurationAnnotation
 
     public function allowArray()
     {
-        return false;
+        return true;
     }
 }
--- a/EventListener/SecurityListener.php
+++ b/EventListener/SecurityListener.php
@@ -48,7 +48,7 @@ class SecurityListener implements EventSubscriberInterface
     public function onKernelController(FilterControllerEvent $event)
     {
         $request = $event->getRequest();
-        if (!$configuration = $request->attributes->get('_security')) {
+        if (!$configurations = $request->attributes->get('_security')) {
             return;
         }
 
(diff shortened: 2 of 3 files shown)


nicolas-grekas:
:+1:

tucksaun:
:+1:

fabpot:
This is a BC break as people might have relied on the fact that a `@Security` annotation on an action overrides one on the class.

hhamon:
Maybe the new behavior should be explicitly and globally configured by the developer under the `sensio_framework_extra` config key.

fabpot:
Please, no global configuration, that never works; what about one bundle needing one value for the global configuration and another one needing another one.

hhamon:
That's a fair point!

iltar:
:+1: from me, but I agree on the BC break, maybe this should cause a major bump.

jvasseur:
A major bump would cause the same problem that a global configuration. Some bundles would need the old version and some the new one.

Maybe create a new Annotation that have the new behavior ?

ghostal:
@iltar directed me to this PR, as I was assuming the `@Security` annotation was "stackable" i.e. could be used on a class, and then applied again on a specific action for further additional restrictions.

I like @jvasseur's suggestion of a new annotation... maybe `@StackableSecurity`?

fabpot:
Can be merged into master after #431 lands.

xabbuh:
#431 was merged.

markyboryslav:
👍

fabpot:
Thank you @lyrixx.

PR #395, opened 2016-01-25
EventListener type check error

RickySu:
following #285 
when setting expression_language to  "security.expression_language" will cause EventListener type check error.

```yaml
#config.yml

sensio_framework_extra:
    security:
        expression_language:  security.expression_language
```

### DIFF
--- a/EventListener/SecurityListener.php
+++ b/EventListener/SecurityListener.php
@@ -11,7 +11,7 @@
 
 namespace Sensio\Bundle\FrameworkExtraBundle\EventListener;
 
-use Sensio\Bundle\FrameworkExtraBundle\Security\ExpressionLanguage;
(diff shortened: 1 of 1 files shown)


fabpot:
You should at least provide the same support as the default expression language, so you should extend `Sensio\Bundle\FrameworkExtraBundle\Security\ExpressionLanguage`

xabbuh:
Maybe a compiler pass checking that the configured service's class is an instance of `Sensio\Bundle\FrameworkExtraBundle\Security\ExpressionLanguage` would be a good idea to detect such mistakes at container compile time.

fabpot:
Closing as there is no feedback. @RickySu Feel free to reopen if you have time to finish it.

PR #398, opened 2016-01-30
Allow to guess templates for controllers in a custom directory

Ener-Getick:
@dunglas recently presented an ADR approach of the controllers with his [ActionBundle](https://github.com/dunglas/DunglasActionBundle).
But ``SensioFrameworkExtraBundle`` can't guess templates path when using it (see https://github.com/dunglas/DunglasActionBundle/issues/5).

So this PR tries to fix this. I'm wondering if this is enough or if we should provide something even more flexible (with a config parameter).

### DIFF
--- a/Templating/TemplateGuesser.php
+++ b/Templating/TemplateGuesser.php
@@ -30,22 +30,31 @@ class TemplateGuesser
     protected $kernel;
 
     /**
+     * @var string[]
+     */
+    private $controllerPatterns;
+
+    /**
      * Constructor.
      *
-     * @param KernelInterface $kernel A KernelInterface instance
+     * @param KernelInterface $kernel             A KernelInterface instance
+     * @param string[]        $controllerPatterns Regexps extracting the controller name from its FQN.
      */
-    public function __construct(KernelInterface $kernel)
+    public function __construct(KernelInterface $kernel, array $controllerPatterns = array())
     {
+        $controllerPatterns[] = '/Controller\\\(.+)Controller$/';
+
         $this->kernel = $kernel;
+        $this->controllerPatterns = $controllerPatterns;
     }
 
     /**
      * Guesses and returns the template name to render based on the controller
      * and action names.
      *
-     * @param array   $controller An array storing the controller object and action method
-     * @param Request $request    A Request instance
-     * @param string  $engine
+     * @param callable $controller An array storing the controller object and action method
+     * @param Request  $request    A Request instance
+     * @param string   $engine
      *
      * @return TemplateReference template reference
      *
@@ -53,13 +62,30 @@ class TemplateGuesser
      */
     public function guessTemplateName($controller, Request $request, $engine = 'twig')
     {
+        if (is_object($controller) && method_exists($controller, '__invoke')) {
+            $controller = array($controller, '__invoke');
+        } elseif (!is_array($controller)) {
+            throw new \InvalidArgumentException(sprintf('First argument of %s must be an array callable or an object defining the magic method __invoke. "%s" given.', __METHOD__, gettype($controller)));
+        }
+
         $className = class_exists('Doctrine\Common\Util\ClassUtils') ? ClassUtils::getClass($controller[0]) : get_class($controller[0]);
 
-        if (!preg_match('/Controller\\\(.+)Controller$/', $className, $matchController)) {
-            throw new \InvalidArgumentException(sprintf('The "%s" class does not look like a controller class (it must be in a "Controller" sub-namespace and the class name must end with "Controller")', get_class($controller[0])));
+        $matchController = null;
+        foreach ($this->controllerPatterns as $pattern) {
+            if (preg_match($pattern, $className, $tempMatch)) {
+                $matchController = $tempMatch;
+                break;
+            }
         }
-        if (!preg_match('/^(.+)Action$/', $controller[1], $matchAction)) {
-            throw new \InvalidArgumentException(sprintf('The "%s" method does not look like an action method (it does not end with Action)', $controller[1]));
+        if (null === $matchController) {
+            throw new \InvalidArgumentException(sprintf('The "%s" class does not look like a controller class (its FQN must match one of the following regexps: "%s")', get_class($controller[0]), implode('", "', $this->controllerPatterns)));
+        }
+
+        if ($controller[1] === '__invoke') {
+            $matchAction = $matchController;
+            $matchController = null;
+        } elseif (!preg_match('/^(.+)Action$/', $controller[1], $matchAction)) {
+            $matchAction = array(null, $controller[1]);
         }
 
         $bundle = $this->getBundleForClass($className);
--- a/Tests/Templating/Fixture/FooBundle/Action/FooAction.php
+++ b/Tests/Templating/Fixture/FooBundle/Action/FooAction.php
@@ -1,0 +1,7 @@
+<?php
+
+namespace Sensio\Bundle\FrameworkExtraBundle\Tests\Templating\Fixture\FooBundle\Action;
+
+class FooAction
+{
+}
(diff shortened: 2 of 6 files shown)


Ener-Getick:
I added a configuration section to define custom patterns.

dunglas:
I was patching DunglasActionBundle to override the template guesser service of this bundle but your approach is more generic.

:+1: for me

bpolaszek:
Definitely +1 for this approach.

Tests/Templating/TemplateGuesserTest.php
```diff
@@ -63,10 +63,10 @@ public function testGuessTemplateNameWithParentBundle()
         $templateGuesser = new TemplateGuesser($this->kernel);
         $templateReference = $templateGuesser->guessTemplateName(array(
             new Fixture\BarBundle\Controller\BarController(),
-            'indexAction',
+            '__invoke',
         ), new Request());
 
-        $this->assertEquals('FooBundle:Bar:index.html.twig', (string) $templateReference);
+        $this->assertEquals('FooBundle::Bar.html.twig', (string) $templateReference);
```
fabpot: You should keep the current test and add a new one instead.

Ener-Getick: done

Templating/TemplateGuesser.php
```diff
@@ -30,13 +30,22 @@ class TemplateGuesser
...
      * Constructor.
      *
-     * @param KernelInterface $kernel A KernelInterface instance
+     * @param KernelInterface $kernel             A KernelInterface instance
+     * @param string[]        $controllerPatterns Regexps extracting the controller name from its FQN.
      */
-    public function __construct(KernelInterface $kernel)
+    public function __construct(KernelInterface $kernel, array $controllerPatterns = array())
     {
+        $controllerPatterns[] = '/Controller\\\(.+)Controller$/';
```
dunglas: What about making this the default value of the config. It will allow to disable template guessing in the `Controller` directory is wanted (more flexible).

Templating/TemplateGuesser.php
```diff
@@ -55,10 +64,21 @@ public function guessTemplateName($controller, Request $request, $engine = 'twig
...
+        }
+        if (null === $matchController) {
+            throw new \InvalidArgumentException(sprintf('The "%s" class does not look like a controller class (its FQN must match one of the following regexps: "%s")', get_class($controller[0]), implode('", "', $this->controllerPatterns)));
         }
-        if (!preg_match('/^(.+)Action$/', $controller[1], $matchAction)) {
+
+        if ($controller[1] === '__invoke') {
+            $matchAction = $matchController;
+            $matchController = null;
+        } elseif (!preg_match('/^(.+)Action$/', $controller[1], $matchAction)) {
```
dunglas: With DunglasActionBundle, the preferred method name is `__invoke` and, as it's recommended to have only 1 action per class, the `Action` suffix is not mandatory in methods name. This test should be removed.

Tests/Templating/TemplateGuesserTest.php
```diff
@@ -103,6 +103,44 @@ public function testGuessTemplateWithoutBundle()
...
+        $this->assertEquals('FooBundle::Foo.html.twig', (string) $templateReference);
+    }
+
+    public function testGuessTemplateWithACustomPattern()
+    {
+        $templateGuesser = new TemplateGuesser($this->kernel, array('/foobar/', '/FooBundle\\\Action\\\(.+)Action/'));
+
+        $templateReference = $templateGuesser->guessTemplateName(array(
+            new Fixture\FooBundle\Action\FooAction(),
+            'indexAction',
```
dunglas: You should add a test with `__invoke()` to prevent regressions.

Ener-Getick: see the test above

Templating/TemplateGuesser.php
Ener-Getick: If the method name match this pattern, the behavior is the same but if not the template name is the full method name.

Tests/Templating/TemplateGuesserTest.php
dunglas: Ok got it.

The is few more cases:

* Implicit reference to `__invoke`: https://github.com/symfony/symfony/blob/master/src/Symfony/Component/HttpKernel/Controller/ControllerResolver.php#L60-L63
* Action method not called `__invoke` and not finishing by `Action`

Ex:

```php
class MyAction()
{
    /** @Template */
    public foo() /* no Action */
    {
    }

    /** @Template */
    public function bar()
    {
    }
}

File: Templating/TemplateGuesser.php (hunk 2 above)
Comment (Ener-Getick) [inline Templating/TemplateGuesser.php]: It would be harder to only add new patterns (which is IMO the major use case)... For example in ``DunglasActionBundle`` you will have to use this in your config to not break existing applications using your bundle.
```yaml
templating:
    controller_patterns: [ /Controller\\(.+)Controller$/, /Action\\(.+)Action$/ ]
```

Moreover this is the last pattern checked so it can easily be overridden.

dunglas: The configuration you show was exactly what I've in mind (it should be the default one).
Most people will never change the default config but some experimented will.

The main benefit is: "I want `@Template` support for new Actions but not for my Legacy controller".

But it's a detail, I personally never use `@Template` so I'll not fight for it :)

Ener-Getick: yeah after all it's not so complicated ^^
I'll make it the default value.

Tests/Templating/TemplateGuesserTest.php
Ener-Getick: Added

dunglas:
Looks good to me

Ener-Getick:
ping @fabpot

Tests/Templating/Fixture/FooBundle/Action/FooAction.php
```diff
@@ -0,0 +1,7 @@
+<?php
+
+namespace Sensio\Bundle\FrameworkExtraBundle\Tests\Templating\Fixture\FooBundle\Action;
+
+class FooAction
+{
+}
```
iltar: Can you add some test cases to verify `@Route` and `@Template` are still working on this for different cases? `__invoke()`, `someAction()`, `fooBar`() etc. The last one is based on @dunglas suggestion regarding the Action suffix.

Ener-Getick: done, I added a ``@dataProvider``.

Ener-Getick:
ping @fabpot is there anything else you want to change ?
It would be great to have this in a stable release.

Templating/TemplateGuesser.php
```diff
@@ -30,22 +30,33 @@ class TemplateGuesser
...
      *
-     * @param KernelInterface $kernel A KernelInterface instance
+     * @param KernelInterface $kernel             A KernelInterface instance
+     * @param string[]        $controllerPatterns Regexps extracting the controller name from its FQN.
      */
-    public function __construct(KernelInterface $kernel)
+    public function __construct(KernelInterface $kernel, array $controllerPatterns = null)
     {
+        if (null === $controllerPatterns) {
+            $controllerPatterns = array('/Controller\\\(.+)Controller$/');
```
fabpot: I think we need to always add this to the list of patterns. If not, as soon as you define the configuration setting, bundles won't work anymore, which would be really hard to understand for the end user. Any downside to always add this?

Ener-Getick: As proposed by @dunglas https://github.com/sensiolabs/SensioFrameworkExtraBundle/pull/398#discussion-diff-51360644 the main benefit to implement it that way is that people can decide to not support their legacy controllers.
Both solutions have their own advantages so I don't know which one we should use...

fabpot: I don't see the downside of keeping the "default" and DX-wise, it's a much better experience.

Ener-Getick: I reverted it to the old behaviour.

fabpot:
Thank you @Ener-Getick.

PR #408, opened 2016-03-22
Check the availibility of Zend Diactoros to auto-enable the PSR7 layer

stof:
As the bundle needs both factories of the bridge, it needs Zend Diactoros to be available too.

See https://github.com/symfony/psr-http-message-bridge/issues/20


dunglas:
:+1:

fabpot:
Thank you @stof.

PR #409, opened 2016-03-22
Guess the template earlier

Ener-Getick:
We have an issue in FOSRestBundle https://github.com/FriendsOfSymfony/FOSRestBundle/issues/1395 since https://github.com/sensiolabs/SensioFrameworkExtraBundle/commit/82b8ad65409c99f693088ecbc09c857f01e3b3fa and https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1355.
So instead of duplicating more logic in FOSRestBundle, IMO the template name should be guessed earlier as it was done before (see [here](https://github.com/sensiolabs/SensioFrameworkExtraBundle/blob/v3.0.12/EventListener/TemplateListener.php#L64)).

### DIFF
--- a/EventListener/TemplateListener.php
+++ b/EventListener/TemplateListener.php
@@ -65,7 +65,13 @@ class TemplateListener implements EventSubscriberInterface
             throw new \InvalidArgumentException('Request attribute "_template" is reserved for @Template annotations.');
         }
 
-        $template->setOwner($event->getController());
+        $template->setOwner($controller = $event->getController());
+
+        // when no template has been given, try to resolve it based on the controller
+        if (null === $template->getTemplate()) {
+            $guesser = $this->container->get('sensio_framework_extra.view.guesser');
+            $template->setTemplate($guesser->guessTemplateName($controller, $request, $template->getEngine()));
+        }
     }
 
     /**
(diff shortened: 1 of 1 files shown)


EventListener/TemplateListener.php
```diff
@@ -88,16 +94,6 @@ public function onKernelView(GetResponseForControllerResultEvent $event)
         $owner = $template->getOwner();
         list($controller, $action) = $owner;
 
-        // when no template has been given, try to resolve it based on the controller
-        if (null === $template->getTemplate()) {
-            if ($action === '__invoke') {
```
Ener-Getick: This is not valid anymore since https://github.com/sensiolabs/SensioFrameworkExtraBundle/pull/398.

fabpot:
LGTM

iltar:
It doesn't really matter in which of the 2 this logic is located for the common use-cases so this looks like a fair change to me.

Now that I think of it (as I'm working on the controller metadata caching), it might be a nice extension point for people to hook into the annotation part of the caching. This information is already available during the cache warm-up and can therefore already calculate the value it should have for the template.

Ener-Getick:
@iltar from what I see many things could be cached so you have my :+1:
Thank you for all the optimizations you made BTW !

fabpot:
Thank you @Ener-Getick.

xabbuh:
In theory this will now break applications where another listener to the `kernel.controller` event changes the controller after this listener has been executed. This is not very likely as the priority here is quite low, but we should keep it in mind.

iltar:
@xabbuh, this is more or less a revert. If a controller changes, there's a lot  more that can break sadly. That's probably why there's so many checks everywhere.

PR #412, opened 2016-04-19
Added regex for uuid 4

Nyholm:
I use an uuid (v4) in my URL to identify an entity (like you always should). This is a recommended security measure if you compare with just using an auto incremental numeric id (which is standard for a Doctrine/Mysql setup). I want of course make sure that my routes are properly defined and that I have a `requirement` on all my variables. So I write something like this:

```php
/**
* @Route("/show/{uuid}", name="show_user", requirements={"uuid" = "[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}"})
*/
public function showAction(User $user)
```

The regex for the uuid is quite long and horrible to write/remember. This PR allows you to write a shortcut for that ugly regex. So instead you would write: 

```php
/**
 * @Route("/show/{uuid}", name="show_user", requirements={"uuid" = "\uuid4"})
 */
public function showAction(User $user)
```


fabpot:
That's nice. I thought about adding such shortcuts a few years ago, but never did it. What about adding this support in the Symfony Routing component instead? We could then add more than just UUID.

Nyholm:
Thank you. I would be happy to make a PR to the routing component. I was not sure if you would accept it there. I'll try to create that PR today.

Nyholm:
The core team did not accept my change. It was too much of a hack. 

However, @lyrixx [provided a neat alternative solution](https://github.com/symfony/symfony/pull/18599#issuecomment-212394889). Do we want to merge this PR or should I update the documentation with the alternative solution?

xabbuh:
@Nyholm FYI, @javiereguiluz started a PR on the docs repo: symfony/symfony-docs#6505

iltar:
If this is added to the Route annotation, it will become quite unclear what the differences are between the symfony route and this extension. As it is, I actually rather see this service removed and make a Controller annotation instead due to this confusion.

fabpot:
I think support for this does not belong to this bundle but to Symfony core. As it has been rejected there, I'm closing this PR.

PR #414, opened 2016-05-21
ParamConverter - Switched options order

JonEastman:
After getting stuck while attempting to use the ParamConverter I realized the docs had the options order backwards.  I used a different example from online with the proposed changes and it works flawlessly. I re-tested it with the current options order that the docs say and it fails again.

P.S. This is my first docs commit so hopefully this isn't just an end-user error


fabpot:
See #783

Issue #338, opened 2015-01-22
Annotations not loaded when defining controller with __invoke() method as a service


gerryvdm:
When defining a controller class with an `__invoke()` method as a service, annotations are not properly loaded when just specifying the service name as the `_controller` route attribute, without specifying the method name. The problem lies in the ControllerListener which first checks if the controller is an array and if not it will not parse the annotations. The reason for this check is that you need to know the class/object and method from which to parse annotations.

Would love to contribute, but not sure where to fix the problem. Could also be solved by changing the ControllerResolver in the FrameworkBundle to return an array of [$service, '__invoke'] instead of just the $service.

iltar:
@gerryvdm do you have an example of the configuration you've used for this?

gerryvdm:
Say I have the following controller class:

    <?php

    use Sensio\Bundle\FrameworkExtraBundle\Configuration\Template;

    class ExampleController
    {
        /**
         * @Template("example.html.twig")
         */
        public function __invoke()
        {
            return array();
        }
    }

My controller service definition:

    services:
        example_controller:
            class: ExampleController

And my routing:

    example_route:
        path: /example
        defaults: { _controller: example_controller }

A request to `/example` will execute the controller, but result in an exception because the @Template annotation is not parsed.

sergiors:
@gerryvdm I think you must have suffix Action

gerryvdm:
@sergiors This is not in fact needed if you have an __invoke() method, see http://symfony.com/doc/current/cookbook/controller/service.html#referring-to-the-service, however, specifying `_controller: example_controller:__invoke` would indeed work and annotations would be loaded, but this is something I would like to prevent.

gerryvdm:
The problem comes on one hand from returning if the controller is not an array here: https://github.com/sensiolabs/SensioFrameworkExtraBundle/blob/master/EventListener/TemplateListener.php#L52-L54, and on the other hand from how the ControllerResolver resolves it here: https://github.com/symfony/symfony/blob/master/src/Symfony/Bundle/FrameworkBundle/Controller/ControllerResolver.php#L66-L67

PR #419, opened 2016-06-29
Change ControllerListener to also load annotations from __invoke

gerryvdm:
For controllers with an __invoke method, annotations are not loaded when not explicitely specifying the method in the _controller attribute.

This change will make sure to load the annotations from the __invoke method.

Fixes https://github.com/sensiolabs/SensioFrameworkExtraBundle/issues/338


fabpot:
Thank you @gerryvdm.

PR #420, opened 2016-07-07
Bug when using @Template in conjunction with @Cache

raphaabreu:
When I use both @Template("...") and @Cache(lastModified="...") every 2 requests I am presented with an error.
It seems that $event->getController() returns a "Closure" object instead of an array and this causes an error to be thrown upon setOwner which takes only arrays.
Adding the check before the setOwner function is called prevents this error.
I wish I could write a test case to show the error but so far I am failing to get phpunit running with the functional test provided.

### DIFF
--- a/EventListener/TemplateListener.php
+++ b/EventListener/TemplateListener.php
@@ -67,7 +67,7 @@ class TemplateListener implements EventSubscriberInterface
 
         $controller = $event->getController();
         if (is_array($controller)) {
-          $template->setOwner($controller);
+            $template->setOwner($controller);
         }
 
         // when no template has been given, try to resolve it based on the controller
--- a/EventListener/TemplateListener.php
+++ b/EventListener/TemplateListener.php
@@ -65,7 +65,10 @@ class TemplateListener implements EventSubscriberInterface
             throw new \InvalidArgumentException('Request attribute "_template" is reserved for @Template annotations.');
         }
 
-        $template->setOwner($controller = $event->getController());
+        $controller = $event->getController();
+        if (is_array($controller)) {
+          $template->setOwner($controller);
+        }
 
         // when no template has been given, try to resolve it based on the controller
         if (null === $template->getTemplate()) {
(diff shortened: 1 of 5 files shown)


iltar:
In here: https://github.com/sensiolabs/SensioFrameworkExtraBundle/blob/master/EventListener/HttpCacheListener.php#L64-L68

It's saying it won't be calling the controller at all if I understand it correctly. This means that there's no reason to even call the `@Template` listener. Shouldn't propagation be stopped instead? ping @fabpot 

The reason this annotation doesn't support Closures, is because closure's can't have the annotations and should therefore not be called.

raphaabreu:
In reading the section you pointed I recalled other issue that might be related: I was also using ETags but the cache was always missing and I noticed that Google Chrome appends "-gzip" to the end of the etag on the request causing it to mismatch. I'll try to write a test case to better illustrate this issue.

raphaabreu:
@iltar, I have submitted a test case that shows what I have been seeing.

It seems that there is a separate issue on the "YamlFileLoader" class that is causing the tests to fail due to a missing YAML::PARSE_CONSTANT in here:
https://github.com/symfony/dependency-injection/blob/678ec7e3cd66cc03382d1b7527804712e29c9504/Loader/YamlFileLoader.php#L370

On my setup I have reverted that file in order to get all tests passing and then did the one I have submitted.

I am sorry if I am being redundant here, I am quite new to contributing to OOS.

iltar:
I still think the event should stop propagation though, now the the template listener is triggered without being used.

snroki:
@raphaabreu to avoid the missing YAML::PARSE_CONSTANT you've to require phpunit with composer directly in the bundle and launch it via ./vendor/bin/phpunit (if you use the .phar file, phpunit will use it's own Yaml loader, cf https://github.com/symfony/symfony/issues/19532 and https://github.com/sebastianbergmann/phpunit/pull/2264)

fabpot:
Closing in favor of #423. 

@raphaabreu If there is another bug, can you open a new PR? Thanks.

PR #422, opened 2016-08-21
[WIP] Adding a smarter AnnotatedRoutingResource class

weaverryan:
Currently, if the file modified time of a controller class is updated, then the cache is marked as stale and all of the routes are rebuilt.

With this addition, a cached resource would only be stale if an annotation has actually changed. This will prevent the route cache from being rebuilt as you're simply building the PHP code in your controller.

Todo
* [ ] Open PR to Symfony's core to support this
* [ ] Add tests (need the above to do that)

### DIFF
--- a/Config/AnnotatedRoutingResource.php
+++ b/Config/AnnotatedRoutingResource.php
@@ -11,17 +11,20 @@
 
 namespace Sensio\Bundle\FrameworkExtraBundle\Config;
 
-class AnnotatedRoutingResource implements \Serializable
+use Symfony\Component\Config\Resource\ResourceInterface;
+use Doctrine\Common\Annotations\Reader;
+
+class AnnotatedRoutingResource implements ResourceInterface, \Serializable
 {
     private $class;
     private $filePath;
     private $annotationMetadata = array();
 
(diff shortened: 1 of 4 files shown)


nicolas-grekas:
Can you please elaborate on why is this an issue actually? I guess building all routes that often degrades DX? By how much? What's the expected factual benefit that justifies the added complexity?

weaverryan:
@nicolas-grekas Fair enough :). It's just a matter of speed: if you use annotation routes, then a high percentage of your dev requests will cause the routing cache to rebuild.

For tiny projects, this isn't noticeable: I just tested and my response time would increase from around 50ms to 90ms. I tested on a bigger, but not huge, project (KnpU) and the difference was about 200-450ms (there's a lot of variance, for some reason). For HUGE projects, annotation routes could become unusable.

So, with that in mind, perhaps this is over-optimizing? It's just been something on my list for a long time, since the cache is unnecessarily built in many cases.

Cheers!

weaverryan:
@nicolas-grekas it would simplify things quite a bit, and hopefully make this feature worth it then :). Thanks for the suggestion - I'll aim to make it happen in core!

Issue #421, opened 2016-07-29
Template::setOwner() must be of the type array, object given


benr77:
SensioFrameworkExtraBundle 3.0.16 with Symfony 2.8.8.

In the production environment, I get the error below when I define caching using the annotation:

`@Cache(lastModified="post.getModified()", ETag="'Post' ~ post.getId() ~ post.getModified().getTimestamp()", public=true)`

And the error is:

`request.CRITICAL: Uncaught PHP Exception Symfony\Component\Debug\Exception\ContextErrorException: "Catchable Fatal Error: Argument 1 passed to Sensio\Bundle\FrameworkExtraBundle\Configuration\Template::setOwner() must be of the type array, object given, called in /app/cache/prod/classes.php on line 8208 and defined" at /vendor/sensio/framework-extra-bundle/Configuration/Template.php line 172`

If I remove the annotation the error goes away.

The error does not occur in the dev environment.

Any ideas?

Thanks

benr77:
This issue appears to completely disappear when reverting to SensioFrameworkExtraBundle 3.0.12

raphaabreu:
I believe that this is related with I have been seeing on #420.

NadaBen:
Hello, did you find a solution for this issue ?

benr77:
I haven't had time to dig any deeper - all I know is that with 3.0.12 it works ok, so I've locked that as my dependency for the time being.

NadaBen:
Okey i will try with 3.0.12 thank you ..

snroki:
Hi, got the same issue with FOSRestBundle when using the @View annotation along with the @Cache (with lastModified or ETag values) and i can't use the 3.0.12 tag because FOSRestBundle 2.0 branch as a minimum of 3.0.13 SensioFrameworkExtraBundle. 
Do you think a fix can be choose ? (Either the check to set the controller only if it's an array or the stopPropagation). Thanks !

NadaBen:
same issue as you @snroki  i'm using FOSRestBundle 2.0 , I can't use the 3.0.12 verion of SensioFrameworkExtraBundle..; i did not find any solution .. :'( still have the same error

PR #423, opened 2016-08-24
Fix error when returning a 304 (not calling template listener) + phpunit

snroki:
Hi,

Since this PR https://github.com/sensiolabs/SensioFrameworkExtraBundle/pull/420 wasn't really accepted i'm submitting another PR with the use of the stopPropagation (works as well).

This should fix this issue : https://github.com/sensiolabs/SensioFrameworkExtraBundle/issues/421

And i fixed an unit test as well which was broken by this commit : https://github.com/symfony/http-foundation/commit/295423690c79e35a59d2aecdc7fb1511171d0c96 (it forces private as default)

### DIFF
--- a/EventListener/HttpCacheListener.php
+++ b/EventListener/HttpCacheListener.php
@@ -65,6 +65,7 @@ class HttpCacheListener implements EventSubscriberInterface
             $event->setController(function () use ($response) {
                 return $response;
             });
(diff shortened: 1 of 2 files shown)


fabpot:
Thank you @snroki.

NadaBen:
.

PR #428, opened 2016-09-12
Use bin/simple-phpunit from phpunit-bridge ~3.2

nicolas-grekas:


fabpot:
Thank you @nicolas-grekas.

Issue #375, opened 2015-10-01
New @CsrfProtected


romaricdrigon:
Hi,

Currently CSRF protection is under-looked in many SF applications, typically with exposed GET requests (deleting an object... pretty much any admin or admin generator!).
That code is required to check a token from a GET parameter:

``` php
        if (!$this->get('security.csrf.token_manager')->isTokenValid(new CsrfToken('intention', $request->query->get('token')))) {
            throw $this->createAccessDeniedException('Invalid CSRF token');
        }
```

I feel like it could be made simpler:

``` php
/**
 * @CsrfProtected("intention")
 */
```

This looks like a small gain, but it make things more "automatic" as security annotation. Pretty much some DX.
I wanted to propose the idea here before working on a PR.

issei-m:
:+1: 
In addition, It'd be more better to get extendability (meaning any HTTP methods/token name are allowed to use), I think.

timglabisch:
:+1:

romqin:
I like the idea. :+1:

hhamon:
Very nice and usefull!

I would name it `CsrfProtection`.

SofHad:
:+1:

Alsatian67:
See #362

romaricdrigon:
Hello @Alsatian67 ,
Then another Exception than AccessDenied should be used, which one though?

apfelbox:
As a side note: meanwhile the main framework has added a helper method to the base controller: https://github.com/symfony/symfony/blob/master/src/Symfony/Bundle/FrameworkBundle/Controller/Controller.php#L436.

```php
$this->isCsrfTokenValid('intention', $request->query->get('token'));
```

So the code without the annotation is a little bit shorter.

Alsatian67:
@romaricdrigon it was just to say that I already proposed some kind of feature ;)

But I think you idea to create another annotation is better than my (provide a helper in the security annotation).

I don't know what to use instead of AccessDenied (for me it doesn't hurt and I don't like to give too much informations).

jameshalsall:
`Symfony\Component\HttpKernel\Exception\BadRequestHttpException`?

fabpot:
Closing as there is a PR about this.

PR #430, opened 2016-09-20
Added CSRF validation (closes #362, #375)

### LINKED ISSUE #362: [DX] Add csrf_provider in @Security annotation ?
Could be like :
```
/*
*@Security("isCsrfTokenValid('secret',csrf)")
*@Route("/message/{id}/remove/{csrf}", name="message_remove")
*/
public function removeMessageAction(Message $message, $csrf){
[...]
}
```

Koc:
Closes #362, #375 

TODO:
- [x] Add tests

### DIFF
--- a/Configuration/Csrf.php
+++ b/Configuration/Csrf.php
@@ -1,0 +1,62 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sensio\Bundle\FrameworkExtraBundle\Configuration;
+
+/**
+ * The Csrf class handles the Csrf annotation.
+ *
+ * @author Konstantin Myakshin <koc-dp@yandex.ru>
+ *
+ * @Annotation
+ * @Target("METHOD")
+ */
+class Csrf extends ConfigurationAnnotation
+{
+    private $intention = 'default';
+
+    private $param = '_token';
+
+    public function getIntention()
+    {
+        return $this->intention;
+    }
+
+    public function setIntention($intention)
+    {
+        $this->intention = $intention;
+    }
+
+    public function getParam()
+    {
+        return $this->param;
+    }
+
+    public function setParam($param)
+    {
+        $this->param = $param;
+    }
+
+    public function setValue($intention)
+    {
+        $this->setIntention($intention);
+    }
+
+    public function getAliasName()
+    {
+        return 'csrf';
+    }
+
+    public function allowArray()
+    {
+        return false;
+    }
+}
--- a/EventListener/CsrfListener.php
+++ b/EventListener/CsrfListener.php
@@ -1,0 +1,62 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sensio\Bundle\FrameworkExtraBundle\EventListener;
+
+use Sensio\Bundle\FrameworkExtraBundle\Configuration\Csrf;
+use Sensio\Bundle\FrameworkExtraBundle\Exception\InvalidCsrfTokenException;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\HttpKernel\Event\FilterControllerEvent;
+use Symfony\Component\HttpKernel\KernelEvents;
+use Symfony\Component\Security\Csrf\CsrfToken;
+use Symfony\Component\Security\Csrf\CsrfTokenManagerInterface;
+
+/**
+ * CsrfListener handles csrf restrictions on actions.
+ *
+ * @author Konstantin Myakshin <koc-dp@yandex.ru>
+ */
+class CsrfListener implements EventSubscriberInterface
+{
+    private $csrfTokenManager;
+
+    public function __construct(CsrfTokenManagerInterface $csrfTokenManager = null)
+    {
+        $this->csrfTokenManager = $csrfTokenManager;
+    }
+
+    public function onKernelController(FilterControllerEvent $event)
+    {
+        $request = $event->getRequest();
+        if (!$configuration = $request->attributes->get('_csrf')) {
+            return;
+        }
+        /* @var $configuration Csrf */
+
+        if (null === $this->csrfTokenManager) {
+            throw new \LogicException('To use the @Csrf tag, you need to install the "symfony/security-csrf".');
+        }
+
+        $tokenValue = $request->get($configuration->getParam());
+
+        if (!$this->csrfTokenManager->isTokenValid(new CsrfToken($configuration->getIntention(), $tokenValue))) {
+            $e = new InvalidCsrfTokenException();
+            $e->setValidToken($this->csrfTokenManager->getToken($configuration->getIntention())->getValue());
+
+            throw $e;
+        }
+    }
+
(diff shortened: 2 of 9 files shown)


Resources/config/csrf.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" ?>
+
+<container xmlns="http://symfony.com/schema/dic/services"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
+
+    <services>
+        <service id="sensio_framework_extra.security.listener" class="Sensio\Bundle\FrameworkExtraBundle\EventListener\CsrfListener">
+            <argument type="service" id="security.csrf.token_manager" on-invalid="null" />
```
Koc: Can I make it required and validate existence when creating new instance?

Koc:
@fabpot tests pass. Please review

EventListener/CsrfListener.php
```diff
@@ -0,0 +1,61 @@
...
+        /* @var $configuration Csrf */
+
+        if (null === $this->csrfTokenManager) {
+            throw new \LogicException('To use the @Csrf tag, you need to install the "symfony/security-csrf".');
+        }
+
+        $token = $request->get($configuration->getParam());
+
+        if (!$this->csrfTokenManager->isTokenValid(new CsrfToken($configuration->getIntention(), $token))) {
+            throw new BadRequestHttpException('Csrf token is invalid.');
```
jakzal: I found examples of both. [Spring returns a 403](http://docs.spring.io/spring-security/site/docs/current/reference/html/csrf.html#csrf-configure):

> The next step is to include Spring Security’s CSRF protection within your application. Some frameworks handle invalid CSRF tokens by invaliding the user’s session, but this causes its own problems. Instead by default Spring Security’s CSRF protection will produce an HTTP 403 access denied.

Configuration/Csrf.php
```diff
@@ -0,0 +1,62 @@
...
+ * @author Konstantin Myakshin <koc-dp@yandex.ru>
+ *
+ * @Annotation
+ * @Target("METHOD")
+ */
+class Csrf extends ConfigurationAnnotation
+{
+    protected $intention;
+
+    protected $param = '_token';
```
fabpot: Can we make the properties private?

EventListener/CsrfListener.php
fabpot: I think I would slightly prefer the AccessDenied one.

Resources/doc/annotations/csrf.rst
```diff
@@ -0,0 +1,38 @@
+@Security
```
fabpot: `@Csrf`?

Resources/doc/annotations/csrf.rst
```diff
@@ -0,0 +1,38 @@
...
+=========
+
+.. caution::
+
+    The ``@Csrf`` annotation only works as of Symfony 2.7.
```
fabpot: `The ``@Csrf`` annotation requires Symfony 2.7+.`?

Resources/doc/annotations/csrf.rst
```diff
@@ -0,0 +1,38 @@
...
+         */
+        public function deleteAction()
+        {
+            // ...
+        }
+    }
+
+On the template side
+
+    <a href="{{ path('acme_post_delete', {'_token': csrf_token('post_delete')}) }}">delete post</a>
```
fabpot: I would prefer that you use an example that does not use GET as it is really a bad practice that we should not encourage.

Tests/EventListener/CsrfListenerTest.php
```diff
@@ -0,0 +1,106 @@
...
+{
+    protected $listener;
+
+    protected $tokenManager;
+
+    protected function setUp()
+    {
+        if (!interface_exists('Symfony\Component\Security\Csrf\CsrfTokenManagerInterface')) {
+            $this->markTestSkipped('This test requires "symfony/security-csrf".');
+        }
```
fabpot: This should be removed as we list all required deps in composer.json so it should always be present.

Tests/EventListener/CsrfListenerTest.php
```diff
@@ -0,0 +1,106 @@
...
+
+        $this->tokenManager = $this
+            ->getMockBuilder('Symfony\Component\Security\Csrf\CsrfTokenManagerInterface')
+            ->getMock();
+
+        $this->listener = new CsrfListener($this->tokenManager);
```
fabpot: Please remove the `setUp()` method and the property and creates instances directly in tests. You can of course create helper methods like you did for the Request and the Event.

Koc:
update:
* annotation properties maked private
* added default intention equal to `default`
* updated docs
* updated tests
* created own exception. I doubt that reusing AccessDeniedException is good idea because of it depends on attributes and subject and it was created for authorization errors. Also this new exception allow pass valid token - useful for handling invalid tokens on client - we can refresh token, ask user and allow retry last action with new token.

Koc:
tests failures looks like unrelated to this PR

Resources/config/csrf.xml
xabbuh: Why not simply removing this definition if the `security.csrf.token_manager` service is not available?

EventListener/CsrfListener.php
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * CsrfListener handles csrf restrictions on actions.
+ *
+ * @author Konstantin Myakshin <koc-dp@yandex.ru>
+ */
+class CsrfListener implements EventSubscriberInterface
+{
+    private $csrfTokenManager;
+
+    public function __construct(CsrfTokenManagerInterface $csrfTokenManager = null)
```
xabbuh: Allowing `null` here makes it possible to create the listener in an unusable state. I think you should rather remove (or not create) the definition instead when no CSRF token manager is available (see also my comment below).

Resources/doc/annotations/csrf.rst
```diff
@@ -0,0 +1,42 @@
...
+         */
+        public function batchDeleteAction()
+        {
+            // ...
+        }
+    }
+
+On the template side
+
+    <form action="{{ path('acme_posts_batch_delete') }}" method="POST">
```
xabbuh: This should be put into a code block too.

EventListener/CsrfListener.php
Koc: I've just reflect logic from security listener.

Resources/doc/annotations/csrf.rst
HeahDude: You can use a double colon for PHP code blocks:
```rst
On the template side::

    <form action="{{ path('acme_posts_batch_delete') }}" method="POST">
```

HeahDude: Edited my post since the code is html :)

Koc:
rebased on master again

ping @fabpot 

questions that required to be resolved:
1a. should I add and register custom exception handler for converting `InvalidCsrfTokenException` to 403 response? Invalid csrf produces 500 error for now. 
1b. AccessDeniedException is not good idea because reasons described in https://github.com/sensiolabs/SensioFrameworkExtraBundle/pull/430#issuecomment-249187438
1c. Or extends from `HttpException` from http-kernel?
2. my listener reflects logic from security listener. But is it possible check availability of the required services inside configuration and enable/disable required annotations depends on services availability. Like https://github.com/symfony/symfony/blob/master/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php#L512 but we need check services.

Koc:
@fabpot can you make review and give advices to my previous questions, please? Would be nice finish this PR.

fabpot:
See #783

PR #431, opened 2016-09-21
Version 4 of the bundle

fabpot:
From the CHANGELOG:

* removed support for the Templating component (only plain Twig is supported)
* removed unneeded phpdocs, converted protected to private properties
* bumped Symfony minimum version to 3.0
* bumped PHP minimum version to 5.5.9
* removed class parameters in container definitions

PR #432, opened 2016-09-22
Adding a new expr option for the Doctrine param converter

weaverryan:
Fulfills #321 

This simply extends the `DoctrineParamConverter`, allowing for this:

```php
/**
 * @Route("/post/{slug}")
 * @ParamConverter("post", options={"expr":"repository.findOneBySlug(slug)"})
 */
```

Of course, `options` is super ugly :). So I've sub-classed `ParamConverter` with a type that simply makes setting the `expr` option much nicer:

```php
/**
 * @Route("/post/{slug}")
 * @Entity("post", expr="repository.findOneBySlug(slug)")
 */
```

All request attributes become variables + `repository` are variables in the expression. No expression functions were added... because I think it's just this simple. All existing functionality is unchanged. In fact, in this example,  you probably don't even need the annotation, since `slug` is probably the field name.

P.S. This is compatible/consistent with the version 4 changes in #431.

Cheers!

### DIFF
--- a/DependencyInjection/SensioFrameworkExtraExtension.php
+++ b/DependencyInjection/SensioFrameworkExtraExtension.php
@@ -41,6 +41,12 @@ class SensioFrameworkExtraExtension extends Extension
         if ($config['router']['annotations']) {
             $annotationsToLoad[] = 'routing.xml';
 
+            if (class_exists('Symfony\Component\ExpressionLanguage\ExpressionLanguage') && class_exists('Symfony\Component\Security\Core\Authorization\ExpressionLanguage')) {
+                $container->setAlias('sensio_framework_extra.converter.doctrine.orm.expression_language', new Alias('sensio_framework_extra.converter.doctrine.orm.expression_language.default', false));
+            } else {
+                $container->removeDefinition('sensio_framework_extra.converter.doctrine.orm.expression_language.default');
+            }
+
             $this->addClassesToCompile(array(
                 'Sensio\\Bundle\\FrameworkExtraBundle\\EventListener\\ControllerListener',
             ));
--- a/DependencyInjection/SensioFrameworkExtraExtension.php
+++ b/DependencyInjection/SensioFrameworkExtraExtension.php
@@ -41,12 +41,6 @@ class SensioFrameworkExtraExtension extends Extension
         if ($config['router']['annotations']) {
             $annotationsToLoad[] = 'routing.xml';
 
-            if (class_exists('Symfony\Component\ExpressionLanguage\ExpressionLanguage') && class_exists('Symfony\Component\Security\Core\Authorization\ExpressionLanguage')) {
-                $container->setAlias('sensio_framework_extra.converter.doctrine.orm.expression_language', new Alias('sensio_framework_extra.converter.doctrine.orm.expression_language.default', false));
-            } else {
-                $container->removeDefinition('sensio_framework_extra.converter.doctrine.orm.expression_language.default');
-            }
-
             $this->addClassesToCompile(array(
                 'Sensio\\Bundle\\FrameworkExtraBundle\\EventListener\\ControllerListener',
             ));
@@ -54,6 +48,12 @@ class SensioFrameworkExtraExtension extends Extension
 
         if ($config['request']['converters']) {
             $annotationsToLoad[] = 'converters.xml';
+
+            if (class_exists('Symfony\Component\ExpressionLanguage\ExpressionLanguage')) {
+                $container->setAlias('sensio_framework_extra.converter.doctrine.orm.expression_language', new Alias('sensio_framework_extra.converter.doctrine.orm.expression_language.default', false));
+            } else {
+                $container->removeDefinition('sensio_framework_extra.converter.doctrine.orm.expression_language.default');
+            }
 
             $this->addClassesToCompile(array(
                 // cannot be added because it has some annotations
--- a/Request/ParamConverter/DoctrineParamConverter.php
+++ b/Request/ParamConverter/DoctrineParamConverter.php
@@ -281,6 +281,15 @@ class DoctrineParamConverter implements ParamConverterInterface
         );
 
         $passedOptions = $configuration->getOptions();
+
+        if (isset($passedOptions['repository_method'])) {
+            @trigger_error('The repository_method option of @ParamConverter is deprecated and will be removed in 5.0. Use the expr option or @Entity.', E_USER_DEPRECATED);
+        }
+
+        if (isset($passedOptions['map_method_signature'])) {
+            @trigger_error('The map_method_signature option of @ParamConverter is deprecated and will be removed in 5.0. Use the expr option or @Entity.', E_USER_DEPRECATED);
+        }
+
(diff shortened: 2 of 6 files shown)


Request/ParamConverter/DoctrineParamConverter.php
```diff
@@ -50,8 +58,16 @@ public function apply(Request $request, ParamConverter $configuration)
...
         }
 
-        // find by identifier?
-        if (false === $object = $this->find($class, $request, $options, $name)) {
+        $errorMessage = null;
+        if ($expr = $options['expr']) {
+            $object = $this->findViaExpression($class, $request, $expr, $options);
+
+            if (null === $object) {
+                $errorMessage = sprintf('The expression "%s" returned null', $expr);
```
weaverryan: This is used below in the `NotFoundHttpException` exception. Do we assume these messages are safe? If not, that's a bummer as it requires us to use really poor exception messages (unless we detected `debug` mode?)

fabpot: Yes, there are safe as they are not displayed anywhere (by default at least). That's also why I don't want to provide translation ou possible customization for them.

Tests/Request/ParamConverter/DoctrineParamConverterTest.php
```diff
@@ -465,4 +473,134 @@ public function testSupportsWithConfiguredEntityManager()
...
+
+        $this->registry->expects($this->once())
+            ->method('getManagerForClass')
+            ->will($this->returnValue($objectManager));
+
+        $this->language->expects($this->once())
+            ->method('evaluate')
+            ->with('repository.findOneByCustomMethod(id)', array(
+                'repository' => $objectRepository,
+                'id' => 5
```
fabpot: missing `,` at the end

fabpot:
By the way, if you want to run the tests locally, you must use `./vendor/bin/simple-phpunit`, not your local phpunit executable.

fabpot:
I haven't had a close looks, but adding expr should probably mean that we can deprecate some other options that are covered by the new ones, right? Or even remove them altogether if we merge this one after the bump to 4.0.

weaverryan:
Thanks for the tip! Tests should be happier now :).

And yes, I think there are a few good options we can deprecate or remove. I've already got work done to remove them locally - I'll try to get to that tomorrow. Then we can decide if we'd rather deprecate.

fabpot:
I would just remove them, and merge after my 4.0 PR.

DependencyInjection/SensioFrameworkExtraExtension.php
```diff
@@ -43,6 +43,12 @@ public function load(array $configs, ContainerBuilder $container)
         if ($config['router']['annotations']) {
             $annotationsToLoad[] = 'routing.xml';
 
+            if (class_exists('Symfony\Component\ExpressionLanguage\ExpressionLanguage') && class_exists('Symfony\Component\Security\Core\Authorization\ExpressionLanguage')) {
+                $container->setAlias('sensio_framework_extra.converter.doctrine.orm.expression_language', new Alias('sensio_framework_extra.converter.doctrine.orm.expression_language.default', false));
+            } else {
```
iltar: While this is correct, you can't test this case. Maybe this is more of a proposition for the DI component though, but it would be really neat if this could be mocked (thus probably wrap the class exists checks)

Request/ParamConverter/DoctrineParamConverter.php
```diff
@@ -202,6 +222,26 @@ private function findDataByMapMethodSignature($em, $class, $repositoryMethod, $c
...
+        }
+
+        $repository = $this->getManager($options['entity_manager'], $class)->getRepository($class);
+        $variables = array_merge($request->attributes->all(), array('repository' => $repository));
+
+        try {
+            return $this->language->evaluate($expression, $variables);
+
+            // execute it here
+        } catch (NoResultException $e) {
```
iltar: What do you mean with execute it here?

iltar:
Maybe a bit off-topic, but are there docs on how to add custom implementations of the ConfigurationAnnotation (+interface)? There's a listener that reads all the annotations and I can swear there were docs of, but I can't find them back...

In theory the above could be combined with your solution here, instead of having to use `options` for everything, create a protected/public API of the `@ParamConverter` to write your own versions. `@DateTime('prop')` would be an option as well.

Request/ParamConverter/DoctrineParamConverter.php
```diff
@@ -229,12 +267,33 @@ public function supports(ParamConverter $configuration)
...
+        }
+
+        if (isset($passedOptions['map_method_signature'])) {
+            @trigger_error('The map_method_signature option of @ParamConverter is deprecated and will be removed in 5.0. Use the expr option or @Entity.', E_USER_DEPRECATED);
+        }
+
+        $extraKeys = array_diff(array_keys($passedOptions), array_keys($defaultValues));
+        if ($extraKeys) {
+            throw new \InvalidArgumentException(sprintf('Invalid option(s) passed to @ParamConverter: %s', implode(', ', $extraKeys)));
+        }
```
weaverryan: I added strict validation on the `options`, which we can do safely now that this class is not extendable (everything is private).

Request/ParamConverter/DoctrineParamConverter.php
```diff
@@ -229,12 +267,33 @@ public function supports(ParamConverter $configuration)
...
+
+        $passedOptions = $configuration->getOptions();
+
+        if (isset($passedOptions['repository_method'])) {
+            @trigger_error('The repository_method option of @ParamConverter is deprecated and will be removed in 5.0. Use the expr option or @Entity.', E_USER_DEPRECATED);
+        }
+
+        if (isset($passedOptions['map_method_signature'])) {
+            @trigger_error('The map_method_signature option of @ParamConverter is deprecated and will be removed in 5.0. Use the expr option or @Entity.', E_USER_DEPRECATED);
+        }
```
weaverryan: We could also consider deprecating `exclude` and maybe `mapping`, though to replace `mapping` with an `expr` is still kind of ugly:

Before:
```php
@Entity("comment", options={"mapping": {"comment_slug": "slug"}})
```

After:
```php
 @Entity("comment", expr="repository.findOneBy({'comment_slug': slug})")
```

Request/ParamConverter/DoctrineParamConverter.php
```diff
@@ -63,7 +79,11 @@ public function apply(Request $request, ParamConverter $configuration)
         }
 
         if (null === $object && false === $configuration->isOptional()) {
-            throw new NotFoundHttpException(sprintf('%s object not found.', $class));
+            $message = sprintf('%s object not found by the @%s annotation.', $class, $this->getAnnotationName($configuration));
```
weaverryan: I tweaked the 404 message to show where this is coming from.

weaverryan:
@iltar I didn't even know that was possible, but yes, now I see it :). Adding (re-adding?) some docs for that would be *awesome* :).

fabpot:
Thank you @weaverryan.

PR #433, opened 2016-09-22
Replaced the PSR7 param converter with argument value resolver

iltar:
Added functional tests to verify the workings, can add unit-tests as well if desired. Functionality for developers is identical to when used with the parameter converter, except that the Symfony Request object no longer has a PSR7 Request/Message object in the attributes.


HeahDude:
Looks good, thanks :) Tests will fail until #431 is merged.

iltar:
@HeahDude yeah, there's some dependency issues that I can't solve until then as well (see php7 failure)

fabpot:
Thank you @iltar.

PR #435, opened 2016-09-23
Adding a new @IsGranted() annotation for checking security

weaverryan:
Since `isGranted()` really should be used (and is in practice) for security in 95%+ of the use-cases, it's just a little bit ugly with the `@Security` expression. This is obviously just a light simplification:

``` php
/**
 * @Route("/post/{slug}")
 *
 * BEFORE
 *
 * @Security("is_granted('ROLE_ADMIN')")
 * @Security("is_granted('EDIT', post)")
 *
 * AFTER
 *
 * @IsGranted("ROLE_ADMIN")
 * @IsGranted("EDIT", object="post")
 */
```

And yes, I'm actually creating an expression inside `@IsGranted`, which works great, but is a bit weird, and doesn't make for the best errors (e.g. if `post` were not a variable). With a bit more work, I could make `@IsGranted` work without needing to have it write an expression.

### DIFF
--- a/Configuration/IsGranted.php
+++ b/Configuration/IsGranted.php
@@ -1,0 +1,107 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sensio\Bundle\FrameworkExtraBundle\Configuration;
+
+/**
+ * The Security class handles the Security annotation.
+ *
+ * @author Ryan Weaver <ryan@knpuniversity.com>
+ * @Annotation
+ */
+class IsGranted extends ConfigurationAnnotation
+{
+    /**
+     * Sets the first argument that will be passed to isGranted().
+     *
+     * @var mixed
+     */
+    private $attributes;
+
+    /**
+     * Sets the second argument passed to isGranted().
+     *
+     * @var mixed
+     */
+    private $subject;
+
+    /**
+     * The message of the exception - has a nice default if not set.
+     *
+     * @var string
+     */
+    private $message;
+
+    /**
+     * If set, will throw Symfony\Component\HttpKernel\Exception\HttpException
+     * with the given $statusCode.
+     * If null, Symfony\Component\Security\Core\Exception\AccessDeniedException.
+     * will be used.
+     *
+     * @var int|null
+     */
+    private $statusCode;
+
+    public function setAttributes($attributes)
+    {
+        $this->attributes = $attributes;
+    }
+
+    public function getAttributes()
+    {
+        return $this->attributes;
+    }
+
+    public function setSubject($subject)
+    {
+        $this->subject = $subject;
+    }
+
+    public function getSubject()
+    {
+        return $this->subject;
+    }
+
+    public function getMessage()
+    {
+        return $this->message;
+    }
+
+    public function setMessage($message)
+    {
+        $this->message = $message;
+    }
+
+    public function getStatusCode()
+    {
+        return $this->statusCode;
+    }
+
+    public function setStatusCode($statusCode)
+    {
+        $this->statusCode = $statusCode;
+    }
+
+    public function setValue($value)
+    {
+        $this->setAttributes($value);
+    }
+
+    public function getAliasName()
+    {
+        return 'is_granted';
+    }
+
+    public function allowArray()
+    {
+        return true;
+    }
+}
--- a/EventListener/IsGrantedListener.php
+++ b/EventListener/IsGrantedListener.php
@@ -1,0 +1,100 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sensio\Bundle\FrameworkExtraBundle\EventListener;
+
+use Sensio\Bundle\FrameworkExtraBundle\Configuration\IsGranted;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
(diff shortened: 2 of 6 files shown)


Configuration/IsGranted.php
```diff
@@ -0,0 +1,68 @@
...
+ * The Security class handles the Security annotation.
+ *
+ * @author Ryan Weaver <ryan@knpuniversity.com>
+ * @Annotation
+ */
+class IsGranted extends Security
+{
+    private $attributes = array();
+
+    private $object;
```
iltar: Should actually be called `subject`. It still specifies `object` in the docs too, but voters already call it a subject: http://api.symfony.com/3.1/Symfony/Component/Security/Core/Authorization/Voter/VoterInterface.html

The fact that the subject is not limited to just objects might help swaying the decision ^^

weaverryan: I agree completely! Updated!

hhamon:
I like this!

However I would just simplify the annotation name to `@Granted` only. What do you think?

iltar:
I think `@IsGranted` is fine, it's the method being executed so it's clear what's being done.

javiereguiluz:
For reference, let's see what other comparable projects do:

### Java

Three annotations cover most cases: `@DenyAll`, `@AllowAll`, `@RolesAllowed`

```java
import javax.annotation.security.*;

@RolesAllowed("RestrictedUsers")
public class Calculator
{
    @RolesAllowed("Administrator")
    public void setNewRate(int rate) {
        //...
    }
    @PermitAll
    public long convertCurrency(long amount) {
        //...
    }
}
```

[Source](http://docs.oracle.com/javaee/7/api/javax/annotation/security/package-summary.html)

### .NET

They just define one annotation (or "attribute", as they call it): `[Authorize]`

```c#
[Authorize]
public class ValuesController : ApiController
{
    [AllowAnonymous]
    public HttpResponseMessage Get() { ... }

    public HttpResponseMessage Post() { ... }
}

[Authorize(Users="Alice,Bob")]
public class ValuesController : ApiController
{
    // ...
}

[Authorize(Roles="Administrators")]
public class ValuesController : ApiController
{
    // ...
}
```

[Source](http://www.asp.net/web-api/overview/security/authentication-and-authorization-in-aspnet-web-api)

---

I'd maintain the `@Security()` annotation and make its behavior be like this one:

 * If you only provide 1 argument, is the required role
 * If you provide 2 arguments, first one is the role and second one is the subject
 * If you want to use an expression, use the named `expr` parameter

```php
/**
 * @Security("ROLE_ADMIN")
 * @Security(roles = "ROLE_ADMIN")  <-- using named arguments
 */


/**
 * @Security("EDIT", "post")
 * @Security(roles = "EDIT", subject = "post")  <-- using named arguments
 */


/**
 * @Security(expr = "is_granted('ROLE_ADMIN') and ...")
 */
```

iltar:
Just please don't call them roles

javiereguiluz:
@iltar you are right, they are not roles. However, I find "attributes" a not so good alternative (I always have problems explaining this concept in Symfony trainings). Would `permission` be a better alternative?

iltar:
@javiereguiluz I never really understood what attributes is supposed to mean so I can understand it's hard to explain, this would be something that should be corrected in Symfony all-together:
 - access_control
 - voters
 - authorization checker
 - access decision manager
 - etc.

Are there cases where you pass along multiple attributes (or what ever the naming will be)? I never really got why it's an array of attributes. I think permission(s) describes what  you are putting in here and would clear up quite a bit in the aforementioned list.

Koc:
@javiereguiluz your proposal is the best. But how can we make it BC? for not if we pass single argument it considered as expression.

fabpot:
4.0 breaks BC anyway. So, documenting the BC break is enough for a PR to be merged in master/4.0.

weaverryan:
I like Javier's ideas too. But, is it possible to pass 2 unnamed args to an annotation?

fabpot:
No, it's not possible.

fabpot:
@weaverryan Do you still want to move forward with this PR? I'm about to release version 4.0.

fabpot:
For the record, I like this PR as is (not necessarily the implementation, but the user facing feature).

weaverryan:
Definitely still interested :). I'm almost done with the "proper" coding of it - I just need to finish the test. I've pushed what I have so far (need to leave right now). But, I'll finish shortly. Thanks for the ping!

weaverryan:
Done! Test failures are some unrelated deprecations.

EventListener/IsGrantedListener.php
```diff
@@ -0,0 +1,100 @@
...
+
+namespace Sensio\Bundle\FrameworkExtraBundle\EventListener;
+
+use Sensio\Bundle\FrameworkExtraBundle\Configuration\IsGranted;
+use Symfony\Component\HttpKernel\Event\FilterControllerEvent;
+use Symfony\Component\HttpKernel\Exception\HttpException;
+use Symfony\Component\HttpKernel\KernelEvents;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\Security\Core\Authorization\AuthorizationCheckerInterface;
+use Symfony\Component\Security\Core\Exception\AccessDeniedException;
```
fabpot: Can you reorder the use statement in alpha order?

weaverryan:
use statements re-ordered

fabpot:
Thank you @weaverryan.

EventListener/IsGrantedListener.php
```diff
@@ -0,0 +1,100 @@
...
+
+        return $argsString;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public static function getSubscribedEvents()
+    {
+        return array(KernelEvents::CONTROLLER => 'onKernelController');
```
stof: this is bad news. It should use the `CONTROLLER_ARGUMENT` event, to be compatible with argument resolvers (here, it would vote on the id of the Post, not on the Post object, when using the Symfony 3.x argument resolvers rather than the ParamConverter layer which has not been migrated to become an argument resolver precisely because of the Security annotation not being migrated to the new event yet)

PR #436, opened 2016-09-23
[WIP] Replace Param converter with configurable argument resolvers

fabpot:
needs symfony/symfony#20036 for more advanced usage.


fabpot:
Some usage examples:

```php
    /**
     * @Route("/posts/{slug}", name="blog_post")
     * @Method("GET")
     * @Arg("post", expr="repository.findOneBySlug(slug)")
     * This @Arg is technically not needed, but here as a usage example
     */
    public function postShowAction(Post $post)
    {
    }
```

```php
    /**
     * @Route("/", defaults={"page": 1}, name="blog_index")
     * @Route("/page/{page}", requirements={"page": "[1-9]\d*"}, name="blog_index_paginated")
     * @Method("GET")
     * @Cache(smaxage="10")
     * @param Post[] $posts
     * @Arg("posts", expr="repository.findLatest(page)")
     */
    public function indexAction($posts, $page)
    {
    }
```

```php
    /**
     * @Route("/{start}/{stop}", defaults={"stop": "2009-09-09"})
     * @Template()
     *
     * @Arg("start", format="Y-m-d:H:i:s")
     * @Arg("stop")
     */
    public function indexAction(\DateTime $start = null, \DateTime $stop)
    {
        return ['start' => $start, 'stop' => $stop];
    }
```

DependencyInjection/SensioFrameworkExtraExtension.php
```diff
@@ -46,24 +46,8 @@ public function load(array $configs, ContainerBuilder $container)
...
-                // cannot be added because it has some annotations
-                //'Sensio\\Bundle\\FrameworkExtraBundle\\Configuration\\ParamConverter',
-                'Sensio\\Bundle\\FrameworkExtraBundle\\EventListener\\ParamConverterListener',
-                'Sensio\\Bundle\\FrameworkExtraBundle\\Request\\ParamConverter\\DateTimeParamConverter',
-                'Sensio\\Bundle\\FrameworkExtraBundle\\Request\\ParamConverter\\DoctrineParamConverter',
-                'Sensio\\Bundle\\FrameworkExtraBundle\\Request\\ParamConverter\\ParamConverterInterface',
-                'Sensio\\Bundle\\FrameworkExtraBundle\\Request\\ParamConverter\\ParamConverterManager',
-            ));
+        if ($config['request']['resolvers']) {
+            $annotationsToLoad[] = 'resolvers.xml';
```
iltar: What if you want to use the `@Arg` feature without using the default set of resolvers? If I'm not using doctrine nor the datetime resolver in a project, I don't really need them but would like to expand on the feature.

Annotation/Arg.php
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * @Annotation
+ * @Target({"METHOD"})
+ * @Attributes({
+ *     @Attribute("value", type="string", required=true),
+ * })
+ */
+class Arg
+{
+    private $arg;
```
iltar: This one is meant to be stand-alone and not a `ConfigurationInterface` implementation?

EventListener/ControllerArgumentOptionsListener.php
```diff
@@ -0,0 +1,138 @@
...
+    {
+        $controller = $event->getController();
+
+        if (!is_array($controller) && method_exists($controller, '__invoke')) {
+            $controller = array($controller, '__invoke');
+        }
+
+        if (!is_array($controller)) {
+            return;
+        }
```
iltar: In theory normal functions can be controllers too (afaik anonymous doesn't support docblocks but not sure), this would make it a string, I think the same goes for static declared methods in a string 'Foo::barAction'.

EventListener/ControllerArgumentOptionsListener.php
```diff
@@ -0,0 +1,138 @@
...
+                continue;
+            }
+
+            try {
+                $options = $this->resolveOptions($event->getRequest(), $parameter->getName(), $argumentMetadata[$parameter->getName()], $options);
+            } catch (OptionsResolverExceptionInterface $e) {
+                throw new LogicException(sprintf('Wrong @Arg configuration for argument "%s" on "%s".', $name, $context), 0, $e);
+            }
+
+            $event->getRequest()->attributes->set($parameter->getName().'_options', $options);
```
iltar: Ah nice leveraging the use of the old setup here, that would prevent them from being executed unless they set a higher priority on their own converter. This should probably be documented as pitfall if they decide to use a higher priority and still try to use `@Arg`.

Request/ArgumentValueResolver/DoctrineResolver.php
```diff
@@ -69,26 +111,26 @@ public function apply(Request $request, ParamConverter $configuration)
             // find by identifier?
         } elseif (false === $object = $this->find($class, $request, $options, $name)) {
             // find by criteria
+// FIXME: to be done elsewhere for expr for instance
+$method = $method->isArray() ? 'findBy' : 'findOneBy';
```
iltar: `$method->isArray()` should probably be `$metadata->isArray()`

I think this case should only occur when guessing, otherwise it's simply a miss-configuration. If it's an array requested, it should probably never even hit `findOne` cases and thus never have an identifier in the URL. That means it can only be `findBy` and deserves its own case in this construction

iltar:
This solution looks a lot more elegant than `ParamConverters` and feels less like a "hack". Is there any specific reason to not have this in Symfony (+doctrine bridge) instead of the bundle at this point? As long as the feature is opt-in it should be all cool.

EventListener/ControllerArgumentOptionsListener.php
fabpot: Anonymous functions do support docblocks, but unfortunately, this is a known limitation of Doctrine Annotations: doctrine/annotations#83

Annotation/Arg.php
fabpot: Yes, it is handled in a total different and more flexible way (options are validated via OptionsResolver). I think `ConfigurationInterface` should die :)

DependencyInjection/SensioFrameworkExtraExtension.php
fabpot: Sure, performance/configurability/lazyness is something that needs to be done. This is an early PoC.

fabpot:
@iltar Having this part of Symfony is indeed a possibility that I've thought of. Creating a new HttpKernel `Annotation` sub-namespace is one idea, creating a new Annotation component is another one. In any case, I'd like this code to be moved outside of a bundle as it can be useful to Drupal or Silex as well.

Annotation/Arg.php
iltar: It's still a pretty cool extension point and used by some packages. Anyway, this is probably a challenge for later on.

https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/master/Configuration/InvalidatePath.php
https://github.com/jaytaph/RateLimitBundle/blob/master/Annotation/RateLimit.php
https://github.com/mayeco/EasyAnnotationBundle

Google `in:github.com "extends ConfigurationAnnotation"`

iltar:
Not just Drupal or Silex, pretty much every framework could use this ^^

Koc:
How we can understand what repository to use when loading array of posts? By parsing param phpdoc?

stof:
This is a huge BC break as long as the SecurityListener is not refactored to run after argument resolvers, as it would mean that the expression does not receive the converted arguments anymore.

EventListener/ControllerArgumentOptionsListener.php
```diff
@@ -0,0 +1,138 @@
...
+
+        if (!is_array($controller)) {
+            return;
+        }
+
+        $className = class_exists('Doctrine\Common\Util\ClassUtils') ? ClassUtils::getClass($controller[0]) : get_class($controller[0]);
+        $method = new \ReflectionMethod($className, $controller[1]);
+        $annotations = $this->reader->getMethodAnnotations($method);
+        $context = $className.'::'.$controller[1].'()';
+        $argumentMetadata = $this->metadataFactory->createArgumentMetadata($controller);
```
iltar: I was just thinking this might need some internal caching so that it's only generated once. Arguments will now be generated in this method *and* the HttpKernel.

Could probably store it in an array internally at least for the current request:
https://github.com/symfony/symfony/blob/master/src/Symfony/Component/HttpKernel/ControllerMetadata/ArgumentMetadataFactory.php

Request/ParamConverter/ParamConverterInterface.php
```diff
@@ -1,40 +0,0 @@
...
-use Sensio\Bundle\FrameworkExtraBundle\Configuration\ParamConverter;
-use Symfony\Component\HttpFoundation\Request;
-
-/**
- * Converts request parameters to objects and stores them as request
- * attributes, so they can be injected as controller method arguments.
- *
- * @author Fabien Potencier <fabien@symfony.com>
- */
-interface ParamConverterInterface
```
HeahDude: We should deprecate it first in 3.x, shouldn't we?

TomasVotruba:
Any ETA?

fabpot:
Closing in favor of symfony/symfony#37829

PR #437, opened 2016-09-29
Don't enable the security listener when security is disabled

### LINKED ISSUE #427: Don't execute listeners for disabled features
In a personal Symfony project I've disabled a lot of features that I don't need, including the entire security system (and the SecurityBundle) and the translation service.

However, SensioFrameworkExtraBundle still calls to some listeners related to Security and Translation:

symfony-timeline

I know that it says `0 ms` time ... but I don't trust that. The only code that takes `0 ms` to execute is the code that it's not executed.

So, should SensioFrameworkExtraBundle check if security is enabled, translation is enabled, etc. before enabling those listeners? Thanks!

javiereguiluz:
This tries to fix #427.

It worked for my application ... but I don't know which is the best way to detect in a bundle extension whether the security component/bundle is enabled or not.

### DIFF
--- a/DependencyInjection/SensioFrameworkExtraExtension.php
+++ b/DependencyInjection/SensioFrameworkExtraExtension.php
@@ -116,6 +116,10 @@ class SensioFrameworkExtraExtension extends Extension
         if ($config['psr_message']['enabled']) {
             $loader->load('psr7.xml');
         }
+
+        if (!$container->hasDefinition('security.token_storage')) {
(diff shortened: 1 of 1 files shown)


HeahDude:
Looks too fragile to me as it depends on the passes order.

fabpot:
To make things lazy-loaded, for this one and to make all the other ones as well, we need a compiler pass to guarantee that everything is loaded and available.

fabpot:
See #441

PR #438, opened 2016-09-29
Removed an unnecesary config file

javiereguiluz:
The `services.xml` doesn't make sense to me ... but maybe I'm missing something here.


fabpot:
Thank you @javiereguiluz.

PR #439, opened 2016-09-29
Don't inject the entire container in the TemplateListener

javiereguiluz:
I don't know if we can make this change ... but injecting the entire container looks unnecessary.

### DIFF
--- a/EventListener/TemplateListener.php
+++ b/EventListener/TemplateListener.php
@@ -95,11 +95,9 @@ class TemplateListener implements EventSubscriberInterface
         }
 
         // attempt to render the actual response
-        $twig = $this->twig;
-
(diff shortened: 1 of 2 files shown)


EventListener/TemplateListener.php
```diff
@@ -94,7 +95,7 @@ public function onKernelView(GetResponseForControllerResultEvent $event)
         }
 
         // attempt to render the actual response
-        $twig = $this->container->get('twig');
+        $twig = $this->twig;
```
fabpot: `$twig` var can be removed an `$this->twig` used instead

fabpot:
apart from the minor comment, 👍

fabpot:
Thank you @javiereguiluz.

PR #441, opened 2016-10-01
optimized listeners by removing unneeded one

### LINKED ISSUE #427: Don't execute listeners for disabled features

fabpot:
Closes #427, replaces #437

### DIFF
--- a/DependencyInjection/Compiler/OptimizerPass.php
+++ b/DependencyInjection/Compiler/OptimizerPass.php
@@ -1,0 +1,34 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sensio\Bundle\FrameworkExtraBundle\DependencyInjection\Compiler;
+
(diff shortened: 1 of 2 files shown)


javiereguiluz:
👍 

I have some questions about other listeners that we could remove:

* `HttpCacheListener` could be disabled when the `sensio_framework_extra.cache.annotations` option is `false`
* `ParamConverterListener` could be disabled when the `sensio_framework_extra.request. converters` option is `false`
* We could do the same in HttpKernel for the TranslatorListener when `framework.translator` is null

fabpot:
@javiereguiluz

 * `HttpCacheListener`: That's already the case. https://github.com/sensiolabs/SensioFrameworkExtraBundle/blob/master/DependencyInjection/SensioFrameworkExtraExtension.php#L70

 * `ParamConverterListener`: Even when `sensio_framework_extra.converter.listener` is `false`, the listener does more things

Am I missing something here?

PR #443, opened 2016-10-11
cache download package archives only

xabbuh:
Caching Packagist metadata will make the cache becoming stale too
often.


fabpot:
Thank you @xabbuh.

PR #445, opened 2016-12-06
[PHP7.1] Allowing nullable types

kubawerlos:
Currently having function in Controller without matching entity like this is fine:
```php
    function myAction(MyEntity $e = null) // ...
```
however this one:
```php
    function myAction(?MyEntity $e) // ...
```
results in exception "MyEntity object not found by the @ParamConverter annotation."

### DIFF
--- a/EventListener/ParamConverterListener.php
+++ b/EventListener/ParamConverterListener.php
@@ -33,6 +33,11 @@ class ParamConverterListener implements EventSubscriberInterface
     protected $autoConvert;
 
     /**
+     * @var bool
+     */
+    private $isParameterTypeSupported;
+
+    /**
      * Constructor.
      *
      * @param ParamConverterManager $manager     A ParamConverterManager instance
@@ -42,6 +47,7 @@ class ParamConverterListener implements EventSubscriberInterface
     {
         $this->manager = $manager;
         $this->autoConvert = $autoConvert;
+        $this->isParameterTypeSupported = method_exists('ReflectionParameter', 'getType');
     }
 
     /**
@@ -96,7 +102,7 @@ class ParamConverterListener implements EventSubscriberInterface
                 $configurations[$name]->setClass($param->getClass()->getName());
             }
 
-            $configurations[$name]->setIsOptional($param->isOptional());
+            $configurations[$name]->setIsOptional($param->isOptional() || $this->isParameterTypeSupported && $param->hasType() && $param->getType()->allowsNull());
         }
 
         return $configurations;
--- a/Tests/EventListener/ParamConverterListenerTest.php
+++ b/Tests/EventListener/ParamConverterListenerTest.php
@@ -13,6 +13,7 @@ namespace Sensio\Bundle\FrameworkExtraBundle\Tests\EventListener;
 
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\ParamConverter;
 use Sensio\Bundle\FrameworkExtraBundle\EventListener\ParamConverterListener;
+use Sensio\Bundle\FrameworkExtraBundle\Tests\EventListener\Fixture\FooControllerNullableParameter;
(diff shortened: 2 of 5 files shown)


EventListener/ParamConverterListener.php
```diff
@@ -96,7 +96,7 @@ private function autoConfigure(\ReflectionFunctionAbstract $r, Request $request,
                 $configurations[$name]->setClass($param->getClass()->getName());
             }
 
-            $configurations[$name]->setIsOptional($param->isOptional());
+            $configurations[$name]->setIsOptional($param->isOptional() || null !== $param->getType() && $param->getType()->allowsNull());
```
pyrech: You should probably check that the `getType` method exist as it does not exist in old version of PHP (this is the reason tests are currently red :wink:)

kubawerlos:
@fabpot any chance for merge?

xabbuh:
@kubawerlos The change looks valid to me, but you should add a test for this if possible.

EventListener/ParamConverterListener.php
```diff
@@ -96,7 +96,7 @@ private function autoConfigure(\ReflectionFunctionAbstract $r, Request $request,
                 $configurations[$name]->setClass($param->getClass()->getName());
             }
 
-            $configurations[$name]->setIsOptional($param->isOptional());
+            $configurations[$name]->setIsOptional($param->isOptional() || method_exists($param, 'hasType') && $param->hasType() && $param->getType()->allowsNull());
```
iltar: Can you add the `method_exists($param, 'hasType')` to the constructor and store it in a property? Something like added here: https://github.com/symfony/symfony/blob/master/src/Symfony/Component/HttpKernel/Controller/ControllerResolver.php#L56

```php
$this->supportsScalarTypes = method_exists('ReflectionParameter', 'getType');
```

Tests/EventListener/ParamConverterListenerTest.php
```diff
@@ -57,6 +57,41 @@ public function testAutoConvert($controllerCallable)
     }
 
     /**
+     * @dataProvider settingOptionalParamProvider
+     * @requires PHP 7.1
+     */
+    public function testSettingOptionalParam($function, $isOptional)
+    {
+        $kernel = $this->getMock('Symfony\Component\HttpKernel\HttpKernelInterface');
+        $request = new Request(array(), array(), array());
```
iltar: If you're passing 3 empty arrays, you might as well omit them as all of them are optional.

iltar:
It would be nice if you could add a few test cases to the functional tests as well, to verify that it integrates nicely in Symfony: https://github.com/sensiolabs/SensioFrameworkExtraBundle/tree/master/Tests/Functional

kubawerlos:
@iltar I'm getting `Symfony\Component\DependencyInjection\Exception\LogicException: Asset support cannot be enabled as the Asset component is not installed.` for every functional test with normal dependencies. With `--prefer-lowest` I'm getting `Class __PHP_Incomplete_Class has no unserializer`.

It suppose to be quick fix for a new functionality, with all these problem I'm giving it up.

iltar:
Hm I don't see why this should be closed this fast. Those issues are probably not even related to your tests

kubawerlos:
Well, I can leave it open until someone will handle Travis tests.

fabpot:
@kubawerlos I've fixed deps in 0c88c0b. If you rebase, I think everything will be ok now.

kubawerlos:
Thanks @fabpot - you have fixed it in master, I have applied these changes to branch 3.0 and it solved CI builds. Nice to see fabbot.io also works properly with nullable types.

fabpot:
Actually, I fixed that in 3.0, not sure why Github thinks that this is on master only. Can you rebase on current 3.0?

fabpot:
Thank you @kubawerlos.

PR #447, opened 2016-12-23
Adding missing use statement for controller.

cmtatro85:
It didn't appear that content was omitted. As a result the code provided would not function.
I understand this example was to show all of the ExtraFramework in action. Perhaps a `// ...` Would of been better than use of the controller.

### DIFF
--- a/Resources/doc/index.rst
+++ b/Resources/doc/index.rst
@@ -125,6 +125,7 @@ The following annotations are defined by the bundle:
 
 This example shows all the available annotations in action::
 
+    use Symfony\Bundle\FrameworkBundle\Controller\Controller;
     use Sensio\Bundle\FrameworkExtraBundle\Configuration\Route;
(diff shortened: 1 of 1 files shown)


Resources/doc/index.rst
```diff
@@ -125,6 +125,7 @@ The following annotations are defined by the bundle:
 
 This example shows all the available annotations in action::
 
+    use Symfony\Bundle\FrameworkBundle\Controller\Controller;
```
fabpot: Actually, I would remove the `extends Controller` part instead as it's not relevant for this example.

xabbuh: Or just add `// ...` to indicate that the example isn't complete (like we do it in the Symfony docs). I mean the namespace declaration is missing too.

fabpot: fixed in 08559c9

PR #449, opened 2017-01-10
Remove usage of addClassesToCompile on PHP7+

fabpot:


jakzal:
Just curious, why is this no longer worth it in PHP 7?

xabbuh:
@jakzal see symfony/symfony#20668 :)

PR #450, opened 2017-01-13
Improve ParamConverter exception messages

apfelbox:
Add the name of the faulty parameter to the exception message.


fabpot:
Thank you @apfelbox.

PR #451, opened 2017-01-26
[Template] Added a way to auto search template in app/ directory

lyrixx:
Symfony best practices state we should use the `app/` folder to store our templates.
So now the `@Template` annotation is not so useful any-more. Let's make it great again ;)

Example in the SE:
```diff
diff --git a/app/Resources/views/default/index.html.twig b/app/Resources/views/Default/index.html.twig
similarity index 100%
rename from app/Resources/views/default/index.html.twig
rename to app/Resources/views/Default/index.html.twig
diff --git a/src/AppBundle/Controller/DefaultController.php b/src/AppBundle/Controller/DefaultController.php
index 5216afe..d984d19 100644
--- a/src/AppBundle/Controller/DefaultController.php
+++ b/src/AppBundle/Controller/DefaultController.php
@@ -3,6 +3,7 @@
 namespace AppBundle\Controller;
 
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Route;
+use Sensio\Bundle\FrameworkExtraBundle\Configuration\Template;
 use Symfony\Bundle\FrameworkBundle\Controller\Controller;
 use Symfony\Component\HttpFoundation\Request;
 
@@ -10,12 +11,13 @@ class DefaultController extends Controller
 {
     /**
      * @Route("/", name="homepage")
+     * @Template(inApp=true)
      */
     public function indexAction(Request $request)
     {
         // replace this example code with whatever you need
-        return $this->render('default/index.html.twig', [
+        return [
             'base_dir' => realpath($this->getParameter('kernel.root_dir').'/..').DIRECTORY_SEPARATOR,
-        ]);
+        ];
     }
 }
```

ping @javiereguiluz

### DIFF
--- a/Configuration/Template.php
+++ b/Configuration/Template.php
@@ -47,6 +47,8 @@ class Template extends ConfigurationAnnotation
      */
     private $owner;
 
+    private $inApp;
+
     /**
      * Returns the array of templates variables.
      *
@@ -153,5 +155,15 @@ class Template extends ConfigurationAnnotation
     public function getOwner()
     {
         return $this->owner;
+    }
+
+    public function getInApp()
+    {
+        return $this->inApp;
+    }
+
+    public function setInApp($inApp)
+    {
+        $this->inApp = $inApp;
     }
 }
--- a/EventListener/TemplateListener.php
+++ b/EventListener/TemplateListener.php
@@ -62,7 +62,7 @@ class TemplateListener implements EventSubscriberInterface
 
         // when no template has been given, try to resolve it based on the controller
         if (null === $template->getTemplate()) {
-            $template->setTemplate($this->templateGuesser->guessTemplateName($controller, $request));
+            $template->setTemplate($this->templateGuesser->guessTemplateName($controller, $request, $template->getInApp()));
         }
     }
 
--- a/Templating/TemplateGuesser.php
+++ b/Templating/TemplateGuesser.php
@@ -54,7 +54,7 @@ class TemplateGuesser
      *
      * @throws \InvalidArgumentException
      */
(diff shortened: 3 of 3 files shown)


Configuration/Template.php
```diff
@@ -47,6 +47,8 @@ class Template extends ConfigurationAnnotation
      */
     private $owner;
 
+    private $inApp;
```
lyrixx: I'm not sure about this naming. The real name should be `bundleLess` but it's too long IMHO.

javiereguiluz:
My question: would it make sense to add this new behavior without a config option?

The idea would be:

1. First look in the bundle's Resources/views/ dir
2. If the template is not found, look for in app/Resources/views/ dir

---

Another question: how is the template transformation in this case?

`@Template` in AppBundle -> DefaultController -> indexAction maps to:

* `default/index.html.twig`
* `app/default/index.html.twig`
* `appbundle/default/index.html.twig`
* ...

lyrixx:
> My question: would it make sense to add this new behavior without a config option?

I like it ;) But it's harder to implement. BTW, I not even sure it's possible

>  AppBundle -> DefaultController -> indexAction maps to:

`Default/index.html.twig` (Lowercase `D`). During my tests, I added `lcfirst`, but IMHO it's not a good idea. It's should be the exact same structure.

fabpot:
I've made a different fix for Flex, which IIUC is exactly what you were trying to do here. See 5e6808f. Basically, if you controller is not in a bundle, it returns no bundle instead of `@Framework` right now. This is consistent with when you do not extends any base controller class.

lyrixx:
I did not test it, but It looks good to me.

PR #454, opened 2017-01-31
param converter priority can be disabled by setting it to boolean false

### LINKED ISSUE #453: disable ParamConverter priority
To disable the ParamConverter priority, you have to specify the string **'false'**. Simply setting boolean false like this does not work:

```
my_converter:
        class: AppBundle\ParamConverter\MyConverter
        tags:
            - { name: request.param_converter, converter: my_converter, priority: false }
```
see [code](https://github.com/sensiolabs/SensioFrameworkExtraBundle/blob/d57c2f297d17ee82baf8cae0b16dae34a9378784/DependencyInjection/Compiler/AddParamConverterPass.php#L38)

Is there any reason not to check for the boolean false?

markusu49:
fixes #453


fabpot:
Thank you @markusu49.

PR #456, opened 2017-02-01
Fix default vars resolution for Template annotation

chalasr:
Fixes broken `@Template` default vars resolution which was relying on a buggy behavior fixed https://github.com/symfony/symfony/pull/21379 (the annotation class loader no longer sets request attributes for any argument that has a default value, but only for ones which may be expected in the route path).


fabpot:
Thank you @chalasr.

PR #459, opened 2017-02-09
Do not throw exception when _template is not an @Template

wouterj:
Currently, when the `_template` request attribute is set, this listener will throw an exception. I don't see why it should throw an exception, returning is enough to not continue.

This will allow other bundles to work with `_template`. For instance, the CmfRoutingBundle uses the `_template` attribute for it's `template_by_type` configuration (which renders a template based on content type). We currently have to do some black magic to rename `_template` to i.e. `contentTemplate` in order to avoid getting this exception.


fabpot:
Thank you @wouterj.

PR #460, opened 2017-02-14
Controller arguments with default values should be default optional

chalasr:
This makes that for e.g:

```php
function fooAction($foo = null, Request $request);
```

the `$foo` parameter is considered optional.

Before it was relying on `ReflectionParameter::isOptional()` only, which returns false if the parameter is not the last. This additionally checks for `isDefaultValueAvailable()`. 
To me it's a bug that was hidden by the fact request attributes were automatically created for any parameter with a default value, that has been fixed in symfony.

Fixes https://github.com/symfony/symfony/issues/21563.

### DIFF
--- a/EventListener/ParamConverterListener.php
+++ b/EventListener/ParamConverterListener.php
@@ -86,23 +86,27 @@ class ParamConverterListener implements EventSubscriberInterface
     private function autoConfigure(\ReflectionFunctionAbstract $r, Request $request, $configurations)
     {
         foreach ($r->getParameters() as $param) {
-            if (!$param->getClass() || $param->getClass()->isInstance($request)) {
+            if ($param->getClass() && $param->getClass()->isInstance($request)) {
                 continue;
             }
 
             $name = $param->getName();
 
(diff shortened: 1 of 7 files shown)


fabpot:
Can you add some tests?

chalasr:
I did, but it required to add doctrine as dev dependency and that seems to reveal an issue in doctrine-bundle's twig requirement...

> Error: Class 'Twig_SimpleFilter' not found
/home/travis/build/sensiolabs/SensioFrameworkExtraBundle/vendor/doctrine/doctrine-bundle/Twig/DoctrineExtension.php:40


See travis build with low deps.

I can't figure it out right now, maybe you have an idea since you made the change that makes use of `Twig_SimpleFilter` in doctrine-bundle?

chalasr:
Changing the twig requirement from `1.11` to `1.12` made it green.
Opened https://github.com/doctrine/DoctrineBundle/pull/614 to fix it on doctrine too.

fabpot:
Thank you @chalasr.

PR #462, opened 2017-02-22
Fix param converter support for nullable scalar/array types

chalasr:
Tests were always broken on my local. That was due to a unregistered bundle (fixture) used by the `NullableAnnotationTest` which requires PHP 7.1, and 7.1 is not enabled on travis.

Enabling the bundle revealed that the test is failing. It expects the following to return `"yes"`:

```php
/**
 * @Route("/nullable")
 */
public function nullableAction(?string $d)
{
    return new Response(null === $d ? 'yes' : 'no');
}
```

But it was throwing instead:

> Controller "NullableArgumentsController" requires that you provide a value for the "$d" argument (because there is no default value or because there is a non optional argument after this one).

Because autoconfiguration is actually skipped if the param doesn't have a class.
This fixes it, enabling 7.1 on travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,7 @@ php:
     - 5.5
     - 5.6
     - 7.0
+    - 7.1
 
 matrix:
     include:
--- a/EventListener/ParamConverterListener.php
+++ b/EventListener/ParamConverterListener.php
@@ -91,21 +91,24 @@ class ParamConverterListener implements EventSubscriberInterface
             }
 
             $name = $param->getName();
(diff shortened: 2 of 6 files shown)


chalasr:
Travis failure is unrelated, it should be broken upstream too (PHPUnit class not found due to the recents changes for phpunit FC).

chalasr:
Ok I think the failure is due to that phpunit-bridge hasn't been released yet. Once done, this actually fixes the bundle test suite.

fabpot:
Thank you @chalasr.

PR #466, opened 2017-03-21
Fix template listener logic

1ed:
The pair of https://github.com/sensiolabs/SensioFrameworkExtraBundle/pull/459

### DIFF
--- a/EventListener/TemplateListener.php
+++ b/EventListener/TemplateListener.php
@@ -55,12 +55,6 @@ class TemplateListener implements EventSubscriberInterface
         $request = $event->getRequest();
         $template = $request->attributes->get('_template');
 
-        // no @Template present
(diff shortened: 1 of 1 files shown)


EventListener/TemplateListener.php
```diff
@@ -55,12 +55,6 @@ public function onKernelController(FilterControllerEvent $event)
         $request = $event->getRequest();
         $template = $request->attributes->get('_template');
 
-        // no @Template present
```
yceruto: Could you keep at least one comment?

1ed: Why is it needed? I think it's pretty obvious what is the intent here.

fabpot:
Thank you @1ed.

PR #471, opened 2017-03-28
Rename property to isOptional to correct autocomplete in annotations plugin

Koc:
IDE suggests annotation fields depends on class properties, so it suggests `optional`. But correct name is `isOptional` because there is setter `setIsOptional`


Koc:
travis fine for now. Could we got this merged?

fabpot:
Thank you @Koc.

PR #473, opened 2017-04-14
Configuration does not override already set response headers

chs2:
Response headers can not be overridden inside a controller if they have been defined through class or method annotations.
```
<?php
namespace AppBundle\Controller;

use AppBundle\Component\Model;
use Symfony\Bundle\FrameworkBundle\Controller\Controller;
use Sensio\Bundle\FrameworkExtraBundle\Configuration\Cache;

/**
 * Class ContentController.
 *
 * @Cache(maxage="60", public=true, smaxage="600")
 */
class ContentController extends Controller
{
    /**
     * @Cache(maxage="60", public=true, smaxage="120")
     */
    public function articleAction(Model $content)
    {
        $response = $this -> render(
            'AppBundle:Content:article.html.twig',
            [
                'content' => $content,
            ]
        );

        if ($content -> getModified() < new \DateTime('6 hours ago')) {
            $response -> setMaxAge(3600);
        }

        return $response;
    }
}

```

### DIFF
--- a/EventListener/HttpCacheListener.php
+++ b/EventListener/HttpCacheListener.php
@@ -94,7 +94,8 @@ class HttpCacheListener implements EventSubscriberInterface
             return;
         }
 
-        if (null !== $age = $configuration->getSMaxAge()) {
+        if (!$response->headers->hasCacheControlDirective('s-maxage')
+            && null !== $age = $configuration->getSMaxAge()) {
             if (!is_numeric($age)) {
                 $now = microtime(true);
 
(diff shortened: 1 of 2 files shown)


iltar:
I think something is wrong with your PR, there are way too many changes in there unrelated to what you're trying to fix. Make sure you pick the same branch to based your changes off as to what you used in git.

chs2:
Right.
I first made it from master to master, then tried to move from 3.0 to 3.0 (don't mind the branch name on my fork).
By the way, which target branch is recommended ?

iltar:
I think the best target is the master, so you should probably base your feature/topic branch on the master as well.

fabpot:
Thank you @chs2.

PR #474, opened 2017-04-21
Remove useless call

Koc:
`$response` is object passed by reference, it can be modified and not requires to be setted again


fabpot:
Thank you @Koc.

PR #475, opened 2017-05-05
Add tip about controller as service with FQCN id

yceruto:


dunglas:
@iltar, yes it works with both.

fabpot:
Thank you @yceruto.

PR #476, opened 2017-05-10
Check that controller is an array before setting it to Template

laurent35240:
Controller returned by FilterControllerEvent is of type callable and not necessary an array.
It can happen that another bundle change this controller and put instead a closure.
We need to check if we can really set this controller to the Template.

It is linked to this issue: https://github.com/sensiolabs/SensioFrameworkExtraBundle/issues/421

### DIFF
--- a/EventListener/TemplateListener.php
+++ b/EventListener/TemplateListener.php
@@ -59,7 +59,10 @@ class TemplateListener implements EventSubscriberInterface
             return;
         }
 
-        $template->setOwner($controller = $event->getController());
+        $controller = $event->getController();
+        if (is_array($controller)) {
+            $template->setOwner($controller);
+        }
 
         // when no template has been given, try to resolve it based on the controller
         if (null === $template->getTemplate()) {
--- a/EventListener/TemplateListener.php
+++ b/EventListener/TemplateListener.php
@@ -88,12 +88,16 @@ class TemplateListener implements EventSubscriberInterface
         }
 
         $parameters = $event->getControllerResult();
-        $owner = $template->getOwner();
-        list($controller, $action) = $owner;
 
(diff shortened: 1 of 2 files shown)


EventListener/TemplateListener.php
```diff
@@ -59,7 +59,10 @@ public function onKernelController(FilterControllerEvent $event)
             return;
         }
 
-        $template->setOwner($controller = $event->getController());
+        $controller = $event->getController();
+        if (is_array($controller)) {
+            $template->setOwner($controller);
```
iltar: Please be ware that this would break `$owner = $template->getOwner();` in `onKernelView`. It should probably do a sanity check now and skip the view listener as the controller has changed.

EventListener/TemplateListener.php
```diff
@@ -88,12 +88,16 @@ public function onKernelView(GetResponseForControllerResultEvent $event)
...
-        $owner = $template->getOwner();
-        list($controller, $action) = $owner;
 
         // when the annotation declares no default vars and the action returns
         // null, all action method arguments are used as default vars
         if (null === $parameters) {
+            $owner = $template->getOwner();
+            $controller = null;
+            $action = null;
+            if ($owner && count($owner) >= 2) {
```
iltar: What about `is_array && is_callable`?

laurent35240: Right it is better, I changed it.

iltar:
There should be full support for tests, would you be able to write those to verify that the changes you've made are working as intended?

laurent35240:
Done.
Unit tests added

fabpot:
If fixed this issue in 9231fff. It should cover all the valid use cases. No need for new tests as the current ones were already broken because of this.

PR #481, opened 2017-05-24
add option "strict" for date creation with DateTime::createFromFormat()

itscaro:
When using format="Y-m-d", the string like "2017-22-05" creates a DateTime of "2018-10-05" with warning.

This case should be considered invalid, for that, an option "strict" is added. When activated, if DateTime::createFromFormat() emits warnings, we will throw Exception too

### DIFF
--- a/Request/ParamConverter/DateTimeParamConverter.php
+++ b/Request/ParamConverter/DateTimeParamConverter.php
@@ -48,6 +48,13 @@ class DateTimeParamConverter implements ParamConverterInterface
         if (isset($options['format'])) {
             $date = $class::createFromFormat($options['format'], $value);
 
+            if (isset($options['strict']) && true === $options['strict']) {
+                $errors = DateTime::getLastErrors();
+                if (0 < $errors['warning_count']) {
+                    $date = false;
+                }
+            }
+
             if (!$date) {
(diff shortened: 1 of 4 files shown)


Request/ParamConverter/DateTimeParamConverter.php
```diff
@@ -46,6 +46,13 @@ public function apply(Request $request, ParamConverter $configuration)
         if (isset($options['format'])) {
             $date = DateTime::createFromFormat($options['format'], $value);
 
+            if (isset($options['strict']) && true === $options['strict']) {
+                $errors = DateTime::getLastErrors();
+                if (0 < $errors['warning_count']) {
+                    $date = false;
+                }
+            }
+
```
nicholasruunu: This whole part could be refactored into the if statement below.

```
isset($options['strict']) && true === $options['strict'] && 0 < DateTime::getLastErrors()['warning_count']
```

Works since 5.4: https://secure.php.net/manual/en/language.types.array.php#example-62

fabpot:
Looks good to me. As this will be merged in master, I would remove the `strict` configuration and make it fail all the time. I don't see how keeping the current behavior would be desirable. You should also add a note in the CHANGELOG to warn users about this change.

PR #482, opened 2017-05-26
Do not overwrite StreamedResponse

slootjes:
The StreamedResponse would be generated correctly when annotated with "isStreamable=true" however it would be overwritten a few lines afterwards with a regular response anyway.

### DIFF
--- a/EventListener/TemplateListener.php
+++ b/EventListener/TemplateListener.php
@@ -103,8 +103,7 @@ class TemplateListener implements EventSubscriberInterface
             };
 
             $event->setResponse(new StreamedResponse($callback));
-        }
-        else {
+        } else {
             $event->setResponse($templating->renderResponse($template->getTemplate(), $parameters));
         }
 
--- a/EventListener/TemplateListener.php
+++ b/EventListener/TemplateListener.php
@@ -104,11 +104,12 @@ class TemplateListener implements EventSubscriberInterface
 
             $event->setResponse(new StreamedResponse($callback));
         }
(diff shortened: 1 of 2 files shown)


EventListener/TemplateListener.php
```diff
@@ -103,12 +103,12 @@ public function onKernelView(GetResponseForControllerResultEvent $event)
             };
 
             $event->setResponse(new StreamedResponse($callback));
+        } else {
+            $event->setResponse($templating->renderResponse($template->getTemplate(), $parameters));
         }
 
         // make sure the owner (controller+dependencies) is not cached or stored elsewhere
         $template->setOwner(array());
```
nicholasruunu: Are you sure this doesn't have to be run before `$event->setResponse()` like before?

slootjes: You either want a streamed response or a regular response. When a streamed response was chosen it was simply overwritten a few lines later on anyway. Putting an else there makes sure it won't overwrite anymore.

nicholasruunu: Yeah I get that, but if you don't want a StreamedResponse shouldn't `$template->setOwner(array());` run before `$event->setResponse($templating->renderResponse($template->getTemplate(), $parameters));`?

slootjes: Afaik the order of where this is placed shouldn't make a difference but please correct me if I'm wrong.

fabpot:
Thank you @slootjes.

PR #488, opened 2017-08-25
Rename services to use their FQCN

fabpot:


DependencyInjection/Compiler/AddExpressionLanguageProvidersPass.php
```diff
@@ -22,8 +22,8 @@ class AddExpressionLanguageProvidersPass implements CompilerPassInterface
      */
     public function process(ContainerBuilder $container)
     {
-        if ($container->hasDefinition('sensio_framework_extra.security.expression_language.default')) {
-            $definition = $container->findDefinition('sensio_framework_extra.security.expression_language.default');
+        if ($container->hasDefinition('Sensio\Bundle\FrameworkExtraBundle\Security\ExpressionLanguage')) {
+            $definition = $container->findDefinition('Sensio\Bundle\FrameworkExtraBundle\Security\ExpressionLanguage');
```
Koc: why not use `class` notation? Anyway you have changed near half of bundle code).

PR #491, opened 2017-08-26
Add must-revalidate support for @Cache annotation

### LINKED ISSUE #243: Feature Request: Must-Revalidate in @Cache
Can we get a must-revalidate option in the @Cache annotation for controllers? What would be the steps involved in adding that?

geoffrey-brier:
This PR fixes #243 
Not sure if I implemented it correctly though.

### DIFF
--- a/Configuration/Cache.php
+++ b/Configuration/Cache.php
@@ -50,6 +50,13 @@ class Cache extends ConfigurationAnnotation
     private $public;
 
     /**
+     * Whether or not the response must be revalidated.
+     *
+     * @var bool
+     */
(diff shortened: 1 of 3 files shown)


fabpot:
Thank you @geoffrey-brier.

magarzon:
There are others Cache-Control parameters that aren't supported either, like no-cache, proxy-revalidate, no-store... I don't know if the best approach it's to have an attribute for each of them in the annotation or to have a Cache-Control attribute that include all of them (after sanitize the values, of course)

PR #492, opened 2017-08-26
Rename ETag to Etag

### LINKED ISSUE #461: Rename ETag to Etag?
In https://github.com/sensiolabs/SensioFrameworkExtraBundle/blob/3.0/Configuration/Cache.php we define `getETag()` and `setETag()` ... and that's why you must use `ETag='...'` in `@Cache()`.

However, in Symfony we use `getEtag()` and `setEtag()` (see https://github.com/symfony/symfony/blob/master/src/Symfony/Component/HttpFoundation/Response.php#L903)

Lastly, the HTTP header is `ETag`.

Should we fix this inconsistency?

geoffrey-brier:
Fixes #461

Wasn't sure if I had to rename all occurences either.

### DIFF
--- a/Configuration/Cache.php
+++ b/Configuration/Cache.php
@@ -207,7 +207,7 @@ class Cache extends ConfigurationAnnotation
      *
      * @param string $expression
      */
-    public function setETag($expression)
+    public function setEtag($expression)
     {
         $this->etag = $expression;
     }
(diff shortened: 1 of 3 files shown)


fabpot:
Thank you @geoffrey-brier.

PR #493, opened 2017-08-26
Fix DoctrineParamConverter (don't call ->find() if it's not the primary key)

### LINKED ISSUE #400: ParamConverter not working for property
Hi, 

https://github.com/sensiolabs/SensioFrameworkExtraBundle/blob/master/Request/ParamConverter/DoctrineParamConverter.php#L54
Isn't this check supposed to be "if (null === $object"?

DoctrineParamConverter::find returns null on NoResultException and the default repository also returns null when no results are found. Why is it checking for false?

geoffrey-brier:
Fixes #400

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ CHANGELOG
 4.0
 ---
 
+ * fixed ParamConverter behaviour to call ->find() only with primary keys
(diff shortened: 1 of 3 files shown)


geoffrey-brier:
@fabpot I was wondering if we should throw a RuntimeException stating that the developer should use the "mapping" option if it's not a primary key ... WDYT ?

fabpot:
See #783

PR #494, opened 2017-08-26
Fix default route names

### LINKED ISSUE #218: Default route name conversion does not follow Symfony defaults
Just an annoying little thing I ran across. The conversion from CamelCase to under_score does not seem to follow Symfony defaults when converting the bundle name for the default routes. For example the `My\TinyCmsBundle` is converted by Symfony itself to `my_tiny_cms`  (for e.g. config variables and other places where the underscored name is used). The default route name generated by the `@Route` annotation for this bundle uses `my_tinycms` instead of `my_tiny_cms`.

It's a small thing but quite annoying, seeing in how many places in Symfony an underscored bundle name is used (config variables, routes, service names, etcetera).

I don't know if this is fixable without breaking BC through... :-/

geoffrey-brier:
Fixes #218

### DIFF
--- a/Routing/AnnotatedRouteControllerLoader.php
+++ b/Routing/AnnotatedRouteControllerLoader.php
@@ -79,8 +79,8 @@ class AnnotatedRouteControllerLoader extends AnnotationClassLoader
         $routeName = parent::getDefaultRouteName($class, $method);
 
         return preg_replace(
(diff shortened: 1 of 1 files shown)


fabpot:
Closing in favor of #497 as it has some unit tests.

Issue #392, opened 2015-12-21
Cache-control: maxStale support


AlmogBaku:
Currently there is no way to define max-stale. (or is it?)

geoffrey-brier:
Not sure if this is an easy pick as the max-stale cache control directive is not currently supported by symfony responses, or am I wrong ? /cc @fabpot

PR #495, opened 2017-08-26
Add cache-control max-stale support (issue #392)

laurent35240:
This PR fixes issue #392

### DIFF
--- a/EventListener/HttpCacheListener.php
+++ b/EventListener/HttpCacheListener.php
@@ -122,7 +122,7 @@ class HttpCacheListener implements EventSubscriberInterface
             if (!is_numeric($stale)) {
                 $now = microtime(true);
 
-                $stale = ceil(strtotime($configuration->getMaxStale(), $now) - $now);
+                $stale = ceil(strtotime($stale, $now) - $now);
             }
 
             $response->headers->addCacheControlDirective('max-stale', $stale);
--- a/EventListener/HttpCacheListener.php
+++ b/EventListener/HttpCacheListener.php
@@ -118,6 +118,16 @@ class HttpCacheListener implements EventSubscriberInterface
             $response->setMaxAge($age);
         }
 
+        if (!$response->headers->hasCacheControlDirective('max-stale') && null !== $stale = $configuration->getMaxStale()) {
+            if (!is_numeric($stale)) {
+                $now = microtime(true);
+
+                $stale = ceil(strtotime($configuration->getMaxStale(), $now) - $now);
+            }
+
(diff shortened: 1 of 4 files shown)


EventListener/HttpCacheListener.php
```diff
@@ -114,6 +114,16 @@ public function onKernelResponse(FilterResponseEvent $event)
             $response->setMaxAge($age);
         }
 
+        if (!$response->headers->hasCacheControlDirective('max-stale') && null !== $stale = $configuration->getMaxStale()) {
+            if (!is_numeric($stale)) {
```
fabpot: In which case wouldn't it be a numeric? `setMaxStale` documents `int`, and nothing in the added docs mention the possibility to pass something else.

EventListener/HttpCacheListener.php
```diff
@@ -114,6 +114,16 @@ public function onKernelResponse(FilterResponseEvent $event)
...
+                $now = microtime(true);
+
+                $stale = ceil(strtotime($configuration->getMaxStale(), $now) - $now);
```
fabpot: Could use `$stale` here instead of `$configuration->getMaxStale()`

laurent35240: I changed it.

EventListener/HttpCacheListener.php
laurent35240: In fact it can be a string with relative time format (1 day, 2 weeks, ...) like for maxage and smaxage.
I updated the phpDoc to clarify this case

fabpot:
Can you rebase to get rid of the merge commit?

Resources/doc/annotations/cache.rst
```diff
@@ -99,6 +99,7 @@ Annotation                                                              Response
 ``@Cache(expires="tomorrow")``                                          ``$response->setExpires()``
 ``@Cache(smaxage="15")``                                                ``$response->setSharedMaxAge()``
 ``@Cache(maxage="15")``                                                 ``$response->setMaxAge()``
+``@Cache(maxstale="15")``                                               ``$response->headers->addCacheControlDirective('max-stale', 15)``
```
fabpot: We should probably add a note before or after this table to explain that one can use text as well.

laurent35240:
Note added in the doc to indicate that we can use string for max, smaxage and maxstale.
I also did some code refactoring on the conversion from string to seconds for these 3 parameters

fabpot:
Thank you @laurent35240.

Issue #283, opened 2014-02-26
Route name prefix?


operandom:
Is it possible to have a name prefix for actions in a controller ?

``` php
/**
 * @Route("/routeprefix", name="routenameprefix_")
 */
class MyController extends Controller
{
}
```

fabpot:
See symfony/symfony#24031

PR #496, opened 2017-08-27
Route name prefix

laurent35240:
For issue #283


pr0nbaer:
Not sure about this.. I guess the actual code change should be done in the Routing component, as the base path of the Route annotation at the class annotation of the controller is handled there as well..

fabpot:
@pr0nbaer indeed, see symfony/symfony#24031

PR #497, opened 2017-08-27
Follow Symfony defaults for routes without name

### LINKED ISSUE #218: Default route name conversion does not follow Symfony defaults

pipo02mix:
minor #218 Default route name conversion does not follow Symfony defaults
This PR was squashed before being merged into the 3.0 branch (closes #218).

Discussion
----------
When a route is defined without name, the default one should follow the Symfony convention
My\TinyCmsBundle -> my_tiny_cms

Commits
-------
[5fef8e4 Follow Symfony defaults for routes without name](https://github.com/sensiolabs/SensioFrameworkExtraBundle/commit/5fef8e4f150f53dd01ffcc9cba137c2eba4910f2)

### DIFF
--- a/Routing/AnnotatedRouteControllerLoader.php
+++ b/Routing/AnnotatedRouteControllerLoader.php
@@ -80,11 +80,12 @@ class AnnotatedRouteControllerLoader extends AnnotationClassLoader
      */
     protected function getDefaultRouteName(\ReflectionClass $class, \ReflectionMethod $method)
     {
-        $routeName = parent::getDefaultRouteName($class, $method);
+        $classMethodParsed = preg_replace('/([a-z])([A-Z])/', '$1_$2', $class->name.'_'.$method->name);
+        $routeName = strtolower(str_replace('\\', '_', $classMethodParsed));
 
         return preg_replace(array(
             '/(bundle|controller)_/',
-            '/action(_\d+)?$/',
+            '/_action(_\d+)?$/',
             '/__/',
         ), array(
             '_',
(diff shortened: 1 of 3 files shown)


Routing/AnnotatedRouteControllerLoader.php
```diff
@@ -80,11 +80,12 @@ protected function getGlobals(\ReflectionClass $class)
      */
     protected function getDefaultRouteName(\ReflectionClass $class, \ReflectionMethod $method)
     {
-        $routeName = parent::getDefaultRouteName($class, $method);
+        $classMethodParsed = preg_replace('/([a-z])([A-Z])/', '$1_$2', $class->name.'_'.$method->name);
+        $routeName = strtolower(str_replace('\\', '_', $classMethodParsed));
```
fabpot: This breaks tests as it does not deal with multi-URLs for a single controller like done in the parent class.

fabpot:
I'm closing as a won't fix as after some investigation, fixing it means overriding the parent methoid from Symfony altogether and so have a different algorithm between Symfony and this bundle. Not worth it if you ask me. Thanks for trying.

PR #311, opened 2014-06-20
Allow custom exception message for @Security annotation

scaytrase:
Custom exception messages are essential to generate customized exception and error page views. Default message is not always acceptable to be displayed to the end user at least because of disclosing security expression used to restrict access.

PR #501, opened 2017-09-12
Allow custom exception message for @Security annotation

scaytrase:
Replaces #311 

> Custom exception messages are essential to generate customized exception and error page views. Default message is not always acceptable to be displayed to the end user at least because of disclosing security expression used to restrict access.

### DIFF
--- a/Resources/doc/annotations/security.rst
+++ b/Resources/doc/annotations/security.rst
@@ -54,6 +54,15 @@ Here is another example, making use of multiple functions in the expression::
     {
     }
 
+The message parameter allows you to customize the text shown to the user at the exception or error pages::
+
+    /**
+     * @Security("is_granted('POST_SHOW', post)", message="You have no rights to view this post.")
+     */
+    public function showAction(Post $post)
+    {
+    }
+
 .. note::
 
     Defining a ``Security`` annotation has the same effect as defining an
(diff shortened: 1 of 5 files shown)


Resources/doc/annotations/security.rst
```diff
@@ -15,7 +15,7 @@ The ``@Security`` annotation restricts access on controllers::
     class PostController extends Controller
     {
         /**
-         * @Security("has_role('ROLE_ADMIN')")
+         * @Security("has_role('ROLE_ADMIN')", message="Staff only.")
```
weaverryan: I'd probably not show the `message` here, so we can show the simplest usage example (and you do a nice job below of showing how to use it).

scaytrase:
Travis failure looks unrelated

Resources/doc/annotations/security.rst
scaytrase: Removed this one

Resources/doc/annotations/security.rst
```diff
@@ -54,6 +54,15 @@ Here is another example, making use of multiple functions in the expression::
     {
     }
 
+The message parameter allows you to customize the text shown to the user at the exception or error pages::
```
fabpot: I would just say: "The message parameter allows you to customize the exception message"

Exception messages are not meant to be displayed to the end user. I understand that it's easier to do it, but still not something I'd recommend.

fabpot:
While reviewing other PRs, I stumbled upon #374, which embeds this change as well. So, I'm going to close this one in favor of it. Thanks for your work on this.

PR #516, opened 2017-10-02
Add autoconfigure tag for ParamConverterInterface

B-Galati:
I tried to add the same behavior in `converters.xml` but it was not working..

```xml
 <instanceof id="Sensio\Bundle\FrameworkExtraBundle\Request\ParamConverter\ParamConverterInterface" public="true">
            <tag name="request.param_converter" />
</instanceof>
```

### DIFF
--- a/Resources/doc/annotations/converters.rst
+++ b/Resources/doc/annotations/converters.rst
@@ -296,7 +296,9 @@ on the request attributes, it should set an attribute named
 ``$configuration->getName()``, which stores an object of class
 ``$configuration->getClass()``.
 
-To register your converter service, you must add a tag to your service:
+If you're using service `auto-registration and autoconfiguration`_,
+you're done! Your converter will automatically be used.
+If not, you must add a tag to your service:
 
 .. configuration-block::
 
@@ -330,3 +332,5 @@ definition.
 .. tip::
 
    Use the ``DoctrineParamConverter`` class as a template for your own converters.
+
+.. _auto-registration and autoconfiguration: http://symfony.com/doc/current/service_container/3.3-di-changes.html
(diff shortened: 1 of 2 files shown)


HeahDude:
>I tried to add the same behavior in converters.xml but it was not working..

It's better to do it in the pass anyway as it allows to auto configure any other converter.

The `instance_of` DI feature only applies on definitions in the same file.

weaverryan:
This makes sense to me: this will create a non-named converter with a priority of 0... which is probably what you want in most cases.

@B-Galati could you also update the docs: https://github.com/sensiolabs/SensioFrameworkExtraBundle/blob/74f1f17e1e5c32ca43de92611e9097a8d0e0acf2/Resources/doc/annotations/converters.rst#creating-a-converter. We should say something like "If you're using service auto-registration and autoconfiguration, you're done! Your converter will automatically be used. If not,..."

You could even link to http://symfony.com/doc/current/service_container/3.3-di-changes.html for `auto-registration and autoconfiguration`.

B-Galati:
@weaverryan It's done thank you, let me know for anything else ;-)

Resources/doc/annotations/converters.rst
```diff
@@ -296,7 +296,11 @@ on the request attributes, it should set an attribute named
 ``$configuration->getName()``, which stores an object of class
 ``$configuration->getClass()``.
 
-To register your converter service, you must add a tag to your service:
+If you're using service `auto-registration and autoconfiguration`_,
+you're done! Your converter will automatically be used.
+If not, you must add a tag to your service:
+
+.. _auto-registration and autoconfiguration: http://symfony.com/doc/current/service_container/3.3-di-changes.html
```
weaverryan: One minor thing, this (by our standard) needs to live all the way at the end of the file :)

B-Galati: done thanks !

fabpot:
Thank you @B-Galati.

PR #517, opened 2017-10-02
Changing how IsGranted works so that it can use all controller args as subject

weaverryan:
This *should* work currently:

```
/**
 * @IsGranted("SOME_ATTR", subject="someVar")
 */
public function fooAction(Request $someVar)
```

The `Request $someVar` is resolved with an argument resolver. It should then be available as a subject to `IsGranted()`. But, this currently does not work: *no* arguments from argument resolvers are available (only `$request->attributes` are used as args).

This has been a long-standing issue with `@Security`. This fixes it for `@IsGranted`. Should we fix this also for `@Security`?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+5.0
+---
+ * Changed the `@Security` annotation to use arguments from argument
+   converters as expression variables.
+
 4.0
 ---
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,10 @@ CHANGELOG
 5.0
 ---
  * Changed the `@Security` annotation to use arguments from argument
-   converters as expression variables.
+   resolvers as expression variables.
+
+ * The `@IsGranted` annotation now also supports using arguments from the
+   argument resolvers as the subject.
 
 4.0
 ---
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ CHANGELOG
 
 5.0
 ---
+
  * Changed the `@Security` annotation to use arguments from argument
    resolvers as expression variables.
 
--- a/EventListener/IsGrantedListener.php
+++ b/EventListener/IsGrantedListener.php
@@ -12,8 +12,8 @@
 namespace Sensio\Bundle\FrameworkExtraBundle\EventListener;
 
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\IsGranted;
+use Sensio\Bundle\FrameworkExtraBundle\Request\ArgumentNameConverter;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
-use Symfony\Component\HttpKernel\ControllerMetadata\ArgumentMetadataFactoryInterface;
 use Symfony\Component\HttpKernel\Event\FilterControllerArgumentsEvent;
 use Symfony\Component\HttpKernel\Exception\HttpException;
 use Symfony\Component\HttpKernel\KernelEvents;
@@ -27,12 +27,12 @@ use Symfony\Component\Security\Core\Exception\AccessDeniedException;
  */
 class IsGrantedListener implements EventSubscriberInterface
 {
-    private $argumentMetadataFactory;
+    private $argumentNameConverter;
     private $authChecker;
 
-    public function __construct(ArgumentMetadataFactoryInterface $argumentMetadataFactory, AuthorizationCheckerInterface $authChecker = null)
+    public function __construct(ArgumentNameConverter $argumentNameConverter, AuthorizationCheckerInterface $authChecker = null)
     {
-        $this->argumentMetadataFactory = $argumentMetadataFactory;
+        $this->argumentNameConverter = $argumentNameConverter;
         $this->authChecker = $authChecker;
     }
 
@@ -48,7 +48,7 @@ class IsGrantedListener implements EventSubscriberInterface
             throw new \LogicException('To use the @IsGranted tag, you need to install symfony/security-bundle and configure your security system.');
         }
 
-        $arguments = $this->getArguments($event);
+        $arguments = $this->argumentNameConverter->getControllerArguments($event);
 
         foreach ($configurations as $configuration) {
             $subject = null;
@@ -72,23 +72,6 @@ class IsGrantedListener implements EventSubscriberInterface
                 throw new AccessDeniedException($message);
             }
         }
-    }
-
-    private function getArguments(FilterControllerArgumentsEvent $event)
-    {
-        $namedArguments = $event->getRequest()->attributes->all();
-        $argumentMetadatas = $this->argumentMetadataFactory->createArgumentMetadata($event->getController());
-
-        // loop over each argument value and its name from the metadata
-        foreach ($event->getArguments() as $index => $argument) {
-            if (!isset($argumentMetadatas[$index])) {
-                throw new \LogicException(sprintf('Could not find any argument metadata for argument %d of the controller.', $index));
-            }
-
-            $namedArguments[$argumentMetadatas[$index]->getName()] = $argument;
-        }
-
-        return $namedArguments;
     }
 
     private function getIsGrantedString(IsGranted $isGranted)
--- a/EventListener/IsGrantedListener.php
+++ b/EventListener/IsGrantedListener.php
@@ -13,7 +13,8 @@ namespace Sensio\Bundle\FrameworkExtraBundle\EventListener;
 
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\IsGranted;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
-use Symfony\Component\HttpKernel\Event\FilterControllerEvent;
+use Symfony\Component\HttpKernel\ControllerMetadata\ArgumentMetadataFactoryInterface;
+use Symfony\Component\HttpKernel\Event\FilterControllerArgumentsEvent;
 use Symfony\Component\HttpKernel\Exception\HttpException;
 use Symfony\Component\HttpKernel\KernelEvents;
 use Symfony\Component\Security\Core\Authorization\AuthorizationCheckerInterface;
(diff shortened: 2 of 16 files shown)


EventListener/IsGrantedListener.php
```diff
@@ -72,6 +75,23 @@ public function onKernelController(FilterControllerEvent $event)
...
+        foreach ($event->getArguments() as $index => $argument) {
+            if (!isset($argumentMetadatas[$index])) {
+                throw new \LogicException(sprintf('Could not find any argument metadata for argument %d of the controller.', $index));
+            }
+
+            $namedArguments[$argumentMetadatas[$index]->getName()] = $argument;
+        }
+
+        return $namedArguments;
+    }
```
weaverryan: This is a technical "bummer". We can use `$event->getArguments()` to get the array of arguments to the controller, but we don't know what *name* of each argument. So, we use the `ArgumentMetadataFactory` to reverse-engineer the argument names. I can't think of a situation where this would *not* work, but it's a bit weird.

stof:
this looks good to me

weaverryan:
Ok, this is ready then!

But what about `@Security`? It is still "un-fixed". We could deprecate `@Security` entirely, fix it (BC break) or just doc it.

xabbuh:
@weaverryan You mean fixing, for example, #514?

weaverryan:
@xabbuh #514 is not exactly what I meant... but actually... yea, it's included. The *real* issue I was talking about is that `@Security` does not take into account the arguments provided by the argument resolver system. So, for example, you cannot use `request` as a variable in `@Security`, because the `Request $request` thing is done with an arg resolver (i.e. there is no `request` in `$request->attributes`. But, about #514, `@Security` also does other things (like always pass a `user` variable that is the currently-authenticated user). So #514 is kind of *another* minor issue with `@Security`. If they used `@IsGranted` instead, then `user` would be the `$user` argument... and never the currently-authenticated user. So, it's at least consistent (it always uses controller args as available subject values).

HeahDude:
So what about https://github.com/sensiolabs/SensioFrameworkExtraBundle/pull/436? Shouldn't this be fixed globally?

weaverryan:
@HeahDude What do you mean fixed globally? Are you thinking of a way to fix `@Security` (i.e. moving it to `CONTROLLER_ARGUMENTS`) without a big BC break?

HeahDude:
Sorry my question is confusing. I'm wondering if that kind of refactoring (#436) would impact what you're trying to solve and if it's not better to consider it before, to make 4.x version of this bundle more consistent?

weaverryan:
Hmm... I think #436 won't change anything. The problem is pretty simple: if we change `@Security` from the `kernel.controller` event to `kernel.controller_arguments`, then it's a weird BC break, as suddenly variables in the expression might change (since arguments made available by argument resolvers are suddenly taken into account). But this *is* the proper behavior. It's a bug fix, that breaks BC in a hard-to-detect kind of way.

#436 changes `@ParamConverter` stuff so that it's done via argument resolvers. If we did that AND changed `@Security` to use the `kernel.controller_arguments` event, it's still the same problem: variables in the expression might change.

So the question is this: if we change the variables in `@Security` to take into account variables/transformations done by argument resolvers, do we consider that a BC break? The new/changed variables would be things like `request`, `session`, `user` (if there is a `UserInterface` type-hinted argument) and anything custom the user added.

stof:
@weaverryan actually, you can access the request with `@security`, because it injects a `request` variable in the expression. But this would not be true for other argument resolvers.

Refactoring the `@Security` annotation would require to keep passing the custom predefined variables (not overwriting them with provided arguments) when a conflict happens. But this is already the case currently in case of conflicts between predefined variables and request attributes.

The main BC break about moving `@Security` to run on the `CONTROLLER_ARGUMENTS` event is that it would make it run later in the process, which can have an impact if you have other event listeners doing another logic on `CONTROLLER` which were set on a lower priority than the SecurityListener and were expecting the Security check to be performed already. But anyway, making IsGranted running on `CONTROLLER_ARGUMENTS` suffers from the same issue (as people switching to the new annotation would delay the security check).
And I'm not sure we have to provide BC there in a major version (we could document it in the UPGRADE file for the few ones relying on this priority). If we consider we have to, we cannot move `IsGranted` to the new event either.

fabpot:
I think it makes sense to fix the issue on @Security as well. We can release a 5.0 even if 4.0 was released a few days ago if that makes everyone less nervous about the changes.

stof:
@weaverryan `request` and `user` would not change in the expression if you keep the merging between predefined variables and argument-based ones in the same order than today.

EventListener/IsGrantedListener.php
```diff
@@ -72,6 +74,23 @@ public function onKernelController(FilterControllerEvent $event)
...
+        $argumentMetadatas = $this->argumentMetadataFactory->createArgumentMetadata($event->getController());
+
+        // loop over each argument value and its name from the metadata
+        foreach ($event->getArguments() as $index => $argument) {
+            if (!isset($argumentMetadatas[$index])) {
+                throw new \LogicException(sprintf('Could not find any argument metadata for argument %d of the controller.', $index));
+            }
+
+            $namedArguments[$argumentMetadatas[$index]->getName()] = $argument;
+        }
```
weaverryan: Is there any issue with variadics here? I'm assuming that if there are 3 arguments to the controller, then `$this->argumentMetadataFactory->createArgumentMetadata()` will always return 3 items exactly.

stof: this is indeed an issue for variadic arguments, as a variadic argument in the metadata will generate multiple values in the final array.

stof: a better solution might be to loop over the metadata instead (and handle the case where the metadata is a variadic argument, by building an array with this argument and all following ones, as done inside the function)

stof:
@weaverryan AFAICT, there would not be much BC break for the change (as long as we keep the merging order) for people not using custom argument resolvers:

- variables defined in https://github.com/sensiolabs/SensioFrameworkExtraBundle/blob/74f1f17e1e5c32ca43de92611e9097a8d0e0acf2/EventListener/SecurityListener.php#L89-L98 would keep being the same
- any request attribute would keep the same by default (as argument resolvers would have read them, and so using this value instead of the request attribute would be the same)

When using custom argument resolvers, there is 2 possible use cases:

- the custom resolver runs at lower priority than the RequestAttributeValueResolver (which runs at priority `100`): no BC break possible, as the custom resolver will only be able to provide arguments which don't exist as request attributes (and so are not available as a variable in the expression in SFEB 3.x or 4.0)
- the custom resolver runs at a high priority (which does not happen in the core): the result of the argument resolver would be used for the expression. I would argue that this would remove a WTF for the user if they were expecting the argument to vote. And this is what would allow us to turn ParamConverter into an ArgumentArgumentResolver in a BC way (as the ParamConverter would stop replacing the request attribute itself, but we want the resolved value to be used in the security check).

CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+5.0
+---
+ * Changed the `@Security` annotation to use arguments from argument
```
stof: you should also mention that `IsGranted` now supports argument resolvers...

weaverryan:
Ok, done!

* This fixes `@Security` so that it also uses argument resolver arguments as variables. But it did not change any merging logic (if a controller argument conflicts with a built-in variable, we still throw an exception)

* I fixed the variadic argument problem with my original PR. The logic is extracted to its own class and tested

* I added a functional test for `@IsGranted` and `@Security`

Request/ArgumentNameConverter.php
```diff
@@ -0,0 +1,61 @@
...
+                break;
+            }
+
+            if (!isset($controllerArguments[$index])) {
+                throw new \LogicException(sprintf('Could not find an argument value for argument %d of the controller.', $index));
+            }
+
+            $namedArguments[$argumentMetadata->getName()] = $controllerArguments[$index];
+            // unset so that we can handle variadic arguments easily
+            unset($controllerArguments[$index]);
```
stof: another option would be to use `array_slice` to handle variadic arguments, which avoids the need to mutate the array here

Resources/config/security.xml
```diff
@@ -17,8 +18,13 @@
         <service id="Sensio\Bundle\FrameworkExtraBundle\Security\ExpressionLanguage" class="Sensio\Bundle\FrameworkExtraBundle\Security\ExpressionLanguage" public="false" />
 
         <service id="Sensio\Bundle\FrameworkExtraBundle\EventListener\IsGrantedListener" class="Sensio\Bundle\FrameworkExtraBundle\EventListener\IsGrantedListener" public="false">
+            <argument type="service" id="Sensio\Bundle\FrameworkExtraBundle\Request\ArgumentNameConverter" />
             <argument type="service" id="security.authorization_checker" on-invalid="null" />
             <tag name="kernel.event_subscriber" />
         </service>
+
+        <service id="Sensio\Bundle\FrameworkExtraBundle\Request\ArgumentNameConverter" class="Sensio\Bundle\FrameworkExtraBundle\Request\ArgumentNameConverter" public="false">
```
stof: no need to specify the class, as it is equal to the id (and Symfony 3.3 is the min version)

Tests/Request/ArgumentNameConverterTest.php
```diff
@@ -0,0 +1,64 @@
...
+
+        // resolves argument names correctly
+        $arg1Metadata = new ArgumentMetadata('arg1Name', 'string', false, false, null);
+        $arg2Metadata = new ArgumentMetadata('arg2Name', 'string', false, false, null);
+        yield array(array('arg1Value', 'arg2Value'), array($arg1Metadata, $arg2Metadata), array('post' => 5), array('post' => 5, 'arg1Name' => 'arg1Value', 'arg2Name' => 'arg2Value'));
+
+        // argument names have priority over request attributes
+        yield array(array('arg1Value', 'arg2Value'), array($arg1Metadata, $arg2Metadata), array('arg1Name' => 'differentValue'), array('arg1Name' => 'arg1Value', 'arg2Name' => 'arg2Value'));
+
+        // variadic arguments are resolved correctly
```
stof: we need 2 cases for the variadic tests:

- variadic argument with args provided
- variadic argument with no arg provided (`arg2ame` will then receive an empty array)`

weaverryan:
Tweaks done! Thanks Stof :)

fabpot:
@stof Can you give your feedback?

Tests/EventListener/IsGrantedListenerTest.php
```diff
@@ -82,18 +100,20 @@ public function testAccessDeniedMessages(array $attributes, $subject, $expectedM
...
-            $request->attributes->set($subject, 'bar');
+            $arguments[$subject] = 'bar';
         }
 
-        $this->setExpectedException(AccessDeniedException::class, $expectedMessage);
+        $listener = new IsGrantedListener($this->createArgumentNameConverter($arguments), $authChecker);
+        $isGranted = new IsGranted(array('attributes' => $attributes, 'subject' => $subject));
+        $request = $this->createRequest($isGranted);
+
+        $this->expectException(AccessDeniedException::class);
```
stof: you need to keep support for PHPUnit 4.8 here, as PHP 5.5 is not supported by newer versions.

fabpot:
I've fixed the tests. Didn't realize it would take me so much time :(

fabpot:
Thank you @weaverryan.

weaverryan:
Bah, sorry about that Fab - didn't see the test failures until now.

PR #518, opened 2017-10-08
Declare the sensio_framework_extra.view.guesser service public

derrabus:
I've tried out one of my projects with Symfony 4 and ran into an issue with SensioFrameworkExtraBundle.

```
Uncaught PHP Exception Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException: "You have requested a non-existent service "sensio_framework_extra.view.guesser"."
```

bildschirmfoto 2017-10-08 um 16 42 26

The problem seems to be that the `TemplateListener` pulls a service out of the container that has not explicitly been declared public. This PR fixes the issue.

### DIFF
--- a/Resources/config/services.xml
+++ b/Resources/config/services.xml
@@ -9,7 +9,7 @@
     </parameters>
 
     <services>
-        <service id="sensio_framework_extra.view.guesser" class="%sensio_framework_extra.view.guesser.class%">
+        <service id="sensio_framework_extra.view.guesser" class="%sensio_framework_extra.view.guesser.class%" public="true">
             <argument type="service" id="kernel" />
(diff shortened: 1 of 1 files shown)


iltar:
For the next version of this bundle, it should probably be made private again and inject the services into the listener

derrabus:
@iltar This is the case in master already. Nevertheless, the 3.0 branch indicates compatibility with Symfony 4, so I thought this issue should be fixed there. :-)

fabpot:
Merging this as a bug fix in 3.0. Probably one of the last release for the 3.0 branch though.

fabpot:
Thank you @derrabus.

PR #520, opened 2017-10-12
Allow nullable argument parameters

benji07:
I just update to the last version and my nullable parameters doesn't work

The following code work with the 4.0 version of the bundle, but not with the 5.0

```
/**
 * @Security("has_role('ROLE_ADMIN')")
 */
public function listAction(User $currentUser, ?Customer $customer = null)
{
}
```


fabpot:
Thank you @benji07.

PR #523, opened 2017-10-20
Fix branch alias.

derrabus:
Currently, master contains the 5.0.x branch. Let's not confuse Composer with outdated information. 😉


fabpot:
Good catch, thanks @derrabus.

PR #529, opened 2017-11-14
Fixed a syntax issue in the docs

javiereguiluz:
This makes docs render wrongly (see https://symfony.com/doc/master/bundles/SensioFrameworkExtraBundle/annotations/converters.html)


fabpot:
Good catch, thanks @javiereguiluz.

PR #530, opened 2017-11-18
Remove internals from autowiring

Simperfit:
The bug does not seems to be in 3.0.

#SymfonyConHackday2017

### DIFF
--- a/DependencyInjection/SensioFrameworkExtraExtension.php
+++ b/DependencyInjection/SensioFrameworkExtraExtension.php
@@ -103,10 +103,10 @@ class SensioFrameworkExtraExtension extends Extension
                 if (class_exists(SecurityExpressionLanguage::class)) {
                     $container->setAlias('sensio_framework_extra.security.expression_language', new Alias($config['security']['expression_language'], false));
                 } else {
-                    $container->removeDefinition('Sensio\Bundle\FrameworkExtraBundle\Security\ExpressionLanguage');
+                    $container->removeDefinition('sensio_framework_extra.security.expression_language.default');
                 }
             } else {
-                $container->removeDefinition('Sensio\Bundle\FrameworkExtraBundle\Security\ExpressionLanguage');
+                $container->removeDefinition('sensio_framework_extra.security.expression_language.default');
             }
 
             if (PHP_VERSION_ID < 70000) {
@@ -137,7 +137,7 @@ class SensioFrameworkExtraExtension extends Extension
 
         if (!empty($config['templating']['controller_patterns'])) {
             $container
-                ->getDefinition('Sensio\Bundle\FrameworkExtraBundle\Templating\TemplateGuesser')
+                ->getDefinition('sensio_framework_extra.view.guesser')
                 ->addArgument($config['templating']['controller_patterns']);
         }
 
--- a/Resources/config/converters.xml
+++ b/Resources/config/converters.xml
@@ -7,22 +7,22 @@
     <services>
         <service id="Sensio\Bundle\FrameworkExtraBundle\EventListener\ParamConverterListener" class="Sensio\Bundle\FrameworkExtraBundle\EventListener\ParamConverterListener" public="false">
             <tag name="kernel.event_subscriber" />
-            <argument type="service" id="Sensio\Bundle\FrameworkExtraBundle\Request\ParamConverter\ParamConverterManager" />
+            <argument type="service" id="sensio_framework_extra.converter.manager" />
             <argument>true</argument>
         </service>
 
-        <service id="Sensio\Bundle\FrameworkExtraBundle\Request\ParamConverter\ParamConverterManager" class="Sensio\Bundle\FrameworkExtraBundle\Request\ParamConverter\ParamConverterManager" />
+        <service id="sensio_framework_extra.converter.manager" class="Sensio\Bundle\FrameworkExtraBundle\Request\ParamConverter\ParamConverterManager" />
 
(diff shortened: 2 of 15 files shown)


DependencyInjection/SensioFrameworkExtraExtension.php
```diff
@@ -103,10 +103,10 @@ public function load(array $configs, ContainerBuilder $container)
                 if (class_exists(SecurityExpressionLanguage::class)) {
                     $container->setAlias('sensio_framework_extra.security.expression_language', new Alias($config['security']['expression_language'], false));
                 } else {
-                    $container->removeDefinition('Sensio\Bundle\FrameworkExtraBundle\Security\ExpressionLanguage');
+                    $container->removeDefinition('framework_extra_bundle.expression_language');
```
nicolas-grekas: the configuration prefix is "sensio_framework_extra", should use the same?

Simperfit:
Yes, I agree with you

2017-11-19 18:26 GMT+01:00 Nicolas Grekas <notifications@github.com>:

nicolas-grekas:
I just found #488, which this PR basically reverts. We should revert to the same names that we used before, don't you think?

Simperfit:
Yeah I agree, I'm going to do that

stof:
Reusing the old name has an advantage: it will avoid breaking bundles needing to deal with these services, if they kept support for the previous version of the bundle (I'm looking at FOSRestBundle here).

fabpot:
Thank you @Simperfit.

Resources/config/converters.xml
```diff
@@ -7,22 +7,22 @@
...
             <argument type="service" id="sensio_framework_extra.converter.doctrine.orm.expression_language" on-invalid="null" />
         </service>
 
-        <service id="Sensio\Bundle\FrameworkExtraBundle\Request\ParamConverter\DateTimeParamConverter" class="Sensio\Bundle\FrameworkExtraBundle\Request\ParamConverter\DateTimeParamConverter" public="false">
+        <service id="framework_extra_bundle.date_time_param_converter" class="Sensio\Bundle\FrameworkExtraBundle\Request\ParamConverter\DateTimeParamConverter" public="false">
             <tag name="request.param_converter" converter="datetime" />
         </service>
 
-        <service id="Symfony\Component\ExpressionLanguage\ExpressionLanguage" class="Symfony\Component\ExpressionLanguage\ExpressionLanguage" public="false" />
+        <service id="framework_extra_bundle.expression_language" class="Symfony\Component\ExpressionLanguage\ExpressionLanguage" public="false" />
```
stof: shouldn't it be `sensio_framework_extra.converter.doctrine.orm.expression_language` to match what is used in the `framework_extra_bundle.doctrine_param_converter` service ?

Resources/config/security.xml
```diff
@@ -5,9 +5,9 @@
     xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
 
     <services>
-        <service id="Sensio\Bundle\FrameworkExtraBundle\EventListener\SecurityListener" class="Sensio\Bundle\FrameworkExtraBundle\EventListener\SecurityListener" public="false">
-            <argument type="service" id="Sensio\Bundle\FrameworkExtraBundle\Request\ArgumentNameConverter" />
-            <argument type="service" id="sensio_framework_extra.security.expression_language" on-invalid="null" />
+        <service id="sensio_framework_extra.security.listener" class="Sensio\Bundle\FrameworkExtraBundle\EventListener\SecurityListener" public="false">
+            <argument type="service" id="framework_extra_bundle.argument_name_convertor" />
+            <argument type="service" id="sensio_framework_extra.security.expression_language.default" on-invalid="null" />
```
stof: that's wrong. It must used the configured service (which will be an alias)

Resources/config/converters.xml
```diff
@@ -7,22 +7,22 @@
...
-            <argument type="service" id="Sensio\Bundle\FrameworkExtraBundle\Request\ParamConverter\ParamConverterManager" />
+            <argument type="service" id="sensio_framework_extra.converter.manager" />
             <argument>true</argument>
         </service>
 
-        <service id="Sensio\Bundle\FrameworkExtraBundle\Request\ParamConverter\ParamConverterManager" class="Sensio\Bundle\FrameworkExtraBundle\Request\ParamConverter\ParamConverterManager" />
+        <service id="sensio_framework_extra.converter.manager" class="Sensio\Bundle\FrameworkExtraBundle\Request\ParamConverter\ParamConverterManager" />
 
-        <service id="Sensio\Bundle\FrameworkExtraBundle\Request\ParamConverter\DoctrineParamConverter" class="Sensio\Bundle\FrameworkExtraBundle\Request\ParamConverter\DoctrineParamConverter" public="false">
+        <service id="framework_extra_bundle.doctrine_param_converter" class="Sensio\Bundle\FrameworkExtraBundle\Request\ParamConverter\DoctrineParamConverter" public="false">
```
stof: Lots of services are missing the `sensio_` part in the prefix, so they don't respect the best practices, and are not reverted to the old names

stof:
@fabpot this PR contains a bunch of mistakes when reverting to old ids

Simperfit:
I will add followup pr with the a bug fix for the mistake.

Ill do that today.

2017-11-30 12:27 GMT+01:00 Christophe Coevoet <notifications@github.com>:

weaverryan:
Yea, I approved too quickly - I thought it was a clean revert of the earlier PR :/

weaverryan:
Fixed in #534

PR #536, opened 2017-11-30
load files before removing definitions

### LINKED ISSUE #532: throw the Fatal error after upgrade v5.1.0
tim 20171130113229

the error:

```
Fatal error: Class 'Symfony\Component\ExpressionLanguage\ExpressionLanguage' not found in /data/www/datitong-admin/vendor/symfony/security/Core/Authorization/ExpressionLanguage.php on line 23

Fatal error: Class 'Symfony\Component\ExpressionLanguage\ExpressionLanguage' not found in /data/www/datitong-admin/vendor/symfony/dependency-injection/ExpressionLanguage.php on line 23
```

I know the reason are ``symfony/expression-language`` missing,  But should not appear after the update.

xabbuh:
Calling `removeDefinition()` before the configuration files have been
loaded may not have the desired effect if the files are only loaded
afterwards.

This should fix #532.


fabpot:
Thank you @xabbuh.

PR #539, opened 2017-12-06
Noted the template/directory change as BC break

### LINKED ISSUE #538: Unable to find template
Since pull #508, we've experteinced the error 'Unable to find template'. It appears [here](https://github.com/sensiolabs/SensioFrameworkExtraBundle/pull/508/files#diff-7cb1d2e06b631ebdc3c5471e12211ff9R70) that `strtolower` is used. Why was this done without notice in the first place?

screenshot from 2017-12-04 14-36-40

iltar:
Fixes #538.

`DefaultIndexController::fooBarAction` produces the following change as of 4.0:
```diff
- DefaultIndex/fooBar.html.twig
+ default_index/foo_bar.html.twig
```

This should've been noted as a BC break when using `@Template()`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,7 @@ CHANGELOG
  * added @IsGranted() annotation
  * allowed to disable some converters
  * allowed to customize the @security message and status code
- * changed template name generation from camelCase to under_score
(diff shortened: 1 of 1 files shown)


fabpot:
Thank you @iltar.

PR #547, opened 2017-12-29
[DoctrineParamConverter] Add evict_cache option

t-veron:
Added an "evict_cache" option to DoctrineParamConverter to force Doctrine to load the entity from the database and not from the cache.


Request/ParamConverter/DoctrineParamConverter.php
```diff
@@ -108,7 +109,15 @@ private function find($class, Request $request, $options, $name)
         } else {
             $method = 'find';
         }
-
+
```
fabpot: extra whitespace

fabpot:
Thank you @t-veron.

PR #549, opened 2018-01-12
Disable Psr7ServerRequestResolver when Diactoros is not installed

### LINKED ISSUE #406: Bundle requires Diactoros even if you only want PSR-7 -> HttpFoundation transformation
I just want to be able to return PSR-7 responses from controller but the implementation in #360 will still load the param converter for HttpFoundation -> PSR-7 which requires Diactoros.

So I get the error 

> Zend Diactoros must be installed to use the DiactorosFactory.

Tobion:
Fixes #406


Tobion:
Failing test unrelated

fabpot:
Thank you @Tobion.

PR #551, opened 2018-02-05
Fix exception with invalid options in DoctrineParamConverter

### LINKED ISSUE #542: Invalid option(s) passed to @ParamConverter: format
After upgrading SF 3.3 to 3.4:

```Invalid option(s) passed to @ParamConverter: format```

Issue is described by https://github.com/kroshp on https://github.com/symfony/symfony/issues/25320
I don't have free time to fully invigistate but probably converter order is invalid - ParamConverterManager should first try apply DateTimeConverter

### PR DESCRIPTION
**Pull request by alcaeus**
Fixes #542 

This happened when we added a param converter for a `DateTime` field with the `format` option but without the `converter` option:
```
@ParamConverter("updatedAt", isOptional=true, options={"format"=DateTime::ATOM})
```

While it may be correct that `DoctrineParamConverter` throws an error on unsupported options, doing so in `supports` is invalid.

### DIFF
--- a/Request/ParamConverter/DoctrineParamConverter.php
+++ b/Request/ParamConverter/DoctrineParamConverter.php
@@ -265,7 +265,7 @@ class DoctrineParamConverter implements ParamConverterInterface
             return false;
         }
 
-        $options = $this->getOptions($configuration);
+        $options = $this->getOptions($configuration, false);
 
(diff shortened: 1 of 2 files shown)


fabpot:
Thank you @alcaeus.

PR #553, opened 2018-02-06
Fix the removal version for the deprecation message

stof:
The 5.x major version is the current one. So removal is for 6.0, not 5.0.


fabpot:
Thank you @stof.

umpirsky:
@stof What will be the alternative? There is nothing about it in [CHANGELOG.md](https://github.com/sensiolabs/SensioFrameworkExtraBundle/blob/master/CHANGELOG.md)?

Is this feature just removed?

stof:
The deprecation message is saying `use the expr option instead`

umpirsky:
I see, it's documented https://github.com/sensiolabs/SensioFrameworkExtraBundle/blob/master/Resources/doc/annotations/converters.rst#2-fetch-via-an-expression. Thanks!

PR #556, opened 2018-02-08
Fix DateTime conversion for optional parameters

alcaeus:
With optional parameters, `DateTimeParamConverter` needs to set the value to `null` and mark the value as converted to avoid later type issues. For example, passing an empty string to a controller with a `DateTimeParamConverter` would cause a type mismatch since the original empty string is not replaced with a `null` value.


fabpot:
Thank you @alcaeus.

PR #557, opened 2018-02-09
Fix tests and add PHP 7.2

alcaeus:


alcaeus:
Note: can't reproduce the deprecations on my machine - anyone have suggestions how to fix?

fabpot:
I've fixed the CS in be47a02

alcaeus:
Thanks @fabpot, I've rebased on top but still need to address the deprecation warnings to get the builds stable.

fabpot:
merging this one as this is still an improvement over the current situation.

fabpot:
Thank you @alcaeus.

PR #559, opened 2018-02-10
Fix `@Method` annotation with several methods

Tobion:
Method::setMethods expects an a single string or an array.


fabpot:
Thank you @Tobion.

PR #560, opened 2018-02-14
Recommend to use Symfony Flex for installation

dunglas:


Resources/doc/index.rst
```diff
@@ -8,7 +8,15 @@ conventions and annotations. It allows for more concise controllers.
 Installation
 ------------
 
-Before using this bundle in your project, add it to your ``composer.json`` file:
+An official `Symfony Flex`_ recipe is available for this bundle. To automatically install and configure it run:
```
fabpot: I try to avoid using Flex. So, I've changed this to `An official Symfony recipe`, which is more than enough.

fabpot:
Thank you @dunglas.

PR #562, opened 2018-02-24
Deprecate routing annotation of FrameworkExtraBundle in favor of Symfony Core

Tobion:
Since https://github.com/symfony/symfony/pull/23044 routing annotation is part of Symfony Core.

The only extra features that the routing annotation from SensioFrameworkExtraBundle has are
- service config for `@Route` which we don't need anymore
    - Either you use class-named services, then you don't need the "service" property at all.
    - Or you don't use class-named services. In this case the better solution to me, is for people to create a alias in the container from the controller class to to service id instead of specifying the service id in the routing. Then you're set as well and that is already common practice with autowiring.
- `@Method` does not provide real value as people agree in https://github.com/symfony/symfony/issues/25103

This resolves https://github.com/symfony/symfony/issues/25103


fabpot:
Thank you @Tobion.

Tests/Fixtures/FooBundle/Controller/InvokableClassLevelController.php
```diff
@@ -11,11 +11,10 @@
 
 namespace Tests\Fixtures\FooBundle\Controller;
 
-use Sensio\Bundle\FrameworkExtraBundle\Configuration\Route;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Template;
+use Symfony\Component\Routing\Annotation\Route;
 
 /**
- * @Route(service="test.invokable_class_level.predefined")
```
stof: wasn't this meant to test the routing loader of this bundle ? Testing the Symfony one does not make sense here.

Tobion: I would assume no, as this test is part of the TemplateAnnotationTest.

iltar: I think I originally created these tests to prevent regressions for the `@Template` annotation. I simply registered them as service this way, so my best guess is that this can be removed.

PR #563, opened 2018-02-28
FIX: We no longer use Foo:Bar:index.html.twig

hipio:
I noticed that the current documentation on the index page, for the `@Template` annotation, is incorrect (since version 4.0), so this PR fixes that.


fabpot:
Thank you @hipio.

PR #567, opened 2018-03-22
Make getIdentifier protected to allow override

mdeboer:
Change the getIdentifier method in the Doctrine param converter from private to protected. This allows overriding the method in classes that extend the param converter. This is useful when you fiddle with
your ID’s in the URL, for example use Hashids.

With the method set to private you can’t override it.

### DIFF
--- a/Request/ParamConverter/DoctrineParamConverter.php
+++ b/Request/ParamConverter/DoctrineParamConverter.php
@@ -125,7 +125,7 @@ class DoctrineParamConverter implements ParamConverterInterface
         }
     }
 
-    private function getIdentifier(Request $request, $options, $name)
+    protected function getIdentifier(Request $request, $options, $name)
     {
         if (null !== $options['id']) {
             if (!is_array($options['id'])) {


iltar:
I'm personally of opinion that you should just write a custom parameter converter (or argument value resolver). There's a lot of voodoo going on in the doctrine parameter converter, that makes it dangerous to use, as it's easy to add unexpected behavior.

mdeboer:
@iltar the only thing the getIdentifier method does is find the ID value from the request so it's harmless and easy to debug (no voodoo there).

I use it to decode [hashed ID's](http://hashids.org) for example. Therefore I call the parent getIdentifier method first and have it extract the ID, then I decode it and return it.

Another case is where you want to extract the ID from the request query parameters instead of route attributes. For example you can do this:

```php
// Check query parameters first
if ($request->query->has($name)) {
    return $request->query->get($name);
}
if ($request->query->has('id') && !$options['id']) {
    return $request->query->get('id');
}

// Fall back to route attributes
return parent::getIdentifier($request, $options, $name);
```

Or check the parent (route attributes) first, whatever you prefer.

There's a lot of things going on in the Doctrine param converter that I don't want to duplicate, I simply want the same functionality but I want to determine what ID is used. Like I said, there's no voodoo here and writing your own would mean duplicating (and maintaining) 99% of the code.

iltar:
In your case, you could also solve it by decoding the parameter and replacing the id value just before the parameter converters run, that way there's no strange customization needed. You could even decorate the service if you wish.

mdeboer:
@iltar that sounds awesome! Can you point me where to look or show me an example?

I had already decorated the service but it still requires this change, no point in having two doctrine converters 😄

iltar:
You could probably do something like this:

```php
final class IdDecodingParameterConverter implements ParamConverterInterface
{
    private $inner;
    private $idDecoder;

    public function __construct(DoctrineParamConverter $inner, IdDecoderInterface $idDecoder)
    {
        $this->inner = $inner;
        $this->idDecoder = $idDecoder;
    }

    public function apply(Request $request, ParamConverter $configuration)
    {
        $this->idDecoder->decodeAttributesInRequest($request, $configuration);

        return $this->inner->apply($request, $configuration);
    }

    public function supports(ParamConverter $configuration)
    {
        return $this->inner->supports($configuration);
    }
}
```

mdeboer:
Ah right, with another param converter. Extending and decorating the doctrine param converter assures that the ID's are being converted, opposed to your example which highly depends on its priority. One mistake in the priorities and it won't run which depending on your caffeine level could take some time to figure out. Also it requires yet another param converter to be initialised instead of extending and decorating it.

Your solution however is more flexible as the `inner` can be any `ParamConverterInterface`.

I'll give it some thought. In the meanwhile I'll leave this open to discuss if you don't mind as it's such a small change without any consequences.

mdeboer:
Okay, that worked and is probably a cleaner and more preferable solution so I'll close this PR. I wrote a ParamConverter that decodes [Hashids](https://hashids.org), I'll open a new PR to include it in the bundle if that's allowed.

PR #571, opened 2018-05-11
Fix unit tests

mdeboer:
I was about to add my own ParamConverter when I noticed the tests were failing and have been failing for quite some time apparently.

I took the liberty to fix those issues:
- Fixed tests marked as risky
- Logging output to file instead of console
- Fixed setExpectedException method not found exception as it's deprecated and removed
- Fixed php-cs-fixer configuration (was missing finder)
- Fixed things found by php-cs-fixer
- Fixed missing fragments renderer
- Fixed missing deprecated LoggerInterface
- Disable deprecation helper by default

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,8 @@
 /vendor
 coverage
 /Tests/Fixtures/cache
+/Tests/Fixtures/logs
 phpunit.xml
 composer.lock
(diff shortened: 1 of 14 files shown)


fabpot:
Thank you @mdeboer.

PR #572, opened 2018-05-12
[ParamConverters] Hashids Converter

mdeboer:
ParamConverter that can decode attributes hashed with [Hashids](https://hashids.org) and pass them on to the inner ParamConverter (e.g. DoctrineParamConverter).

As the inner ParamConverter is free to choose I did not include a service definition. To use it you’ll need to add a service definition and set a priority (higher than the inner).

Also see closed PR #567

### DIFF
--- a/Request/ParamConverter/HashidsParamConverter.php
+++ b/Request/ParamConverter/HashidsParamConverter.php
@@ -1,0 +1,98 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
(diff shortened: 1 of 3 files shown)


mdeboer:
Hang on, I see my previous PR was merged. I'll update this PR with latest master.

mdeboer:
Okay, things are fixed except for PHP <= 5.5 support as Hashids requires >= 5.5. Maybe add the `--ignore-platform-reqs` parameter to the `composer install` command in travis? Then I'll only need to add the php requirements in the tests for it to work.

fabpot:
I don't think this belongs to this bundle. I don't want to add support for more converters if they are not widely useful. Thanks for proposing.

PR #576, opened 2018-06-29
Make "sensio_framework_extra.router.annotations" disabled by default

Mediagone:
The config builder was not updated to match the new deprecations introduced in 5.2, then having `sensio_framework_extra.router.annotations` still defaulted to TRUE raises deprecation notices if no `framework_extra.yaml` file (with annotations disabled) is present in the project.

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -37,7 +37,7 @@ class Configuration implements ConfigurationInterface
                 ->arrayNode('router')
                     ->addDefaultsIfNotSet()
                     ->children()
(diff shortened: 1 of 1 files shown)


cybernet:
and this isn't merged ... because ?

xabbuh:
Changing the default value would be a BC break. But there is a pull request for the recipe (see symfony/recipes#430) to disable this option by default when the bundle is installed in a Flex application.

fabpot:
Closing as we cannot change the default value.

fabpot:
I've done the change directly in the Symfony Standard Edition: see symfony/symfony-standard#1133

PR #579, opened 2018-07-19
Added failing test case for template guesser bug

Padam87:
In a Symfony 4, if you create a controller in `App\Controller` which extends a controller in a bundle, the `@Template` annotation will try to render the wrong view.

The main controller is not in a bundle, so the guesser will continue with the parent controller, and wrongly guess the parent's bundle.


Padam87:
Not sure how to go about fixing this.

Maybe check if the class is in the same namespace as the kernel?

fabpot:
See #783

Issue #580, opened 2018-07-20
Doctrine/common is deprecated


MaxandreOgeret:
https://github.com/doctrine/common/issues/826
https://github.com/symfony/symfony/pull/27609

nicolas-grekas:
Would you like to submit a PR?
You could take inspiration from https://github.com/symfony/symfony/pull/28038

PR #582, opened 2018-07-27
Remove direct dependencies on doctrine/common

MaxandreOgeret:
Removing deprecated doctrine/common.

https://github.com/sensiolabs/SensioFrameworkExtraBundle/issues/580#issuecomment-408376379

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,9 @@
         "symfony/framework-bundle": "^3.4|^4.0",
         "symfony/dependency-injection": "^3.3|^4.0",
         "symfony/http-kernel": "^3.3|^4.0",
-        "doctrine/common": "^2.2"
+        "doctrine/collections": "~1.0",
+        "doctrine/event-manager": "~1.0",
+        "doctrine/persistence": "~1.0"
     },
     "require-dev": {
         "symfony/expression-language": "^3.3|^4.0",
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,15 +8,12 @@ cache:
         - .phpunit
 
 php:
-    - 5.5
-    - 5.6
-    - 7.0
     - 7.1
     - 7.2
 
 matrix:
     include:
-        - php: 7.0
+        - php: 7.1
           env: deps=low
     fast_finish: true
 


MaxandreOgeret:
I guess  
"doctrine/collections"
"doctrine/event-manager"
"doctrine/persistence":
are only compatible with php version >= 7.1

nicolas-grekas:
Yep, this forces PHP7.1 or higher. We'll have to bump the major version here to do this change.
The Travis matrix needs to be updated to make tests pass of course.

MaxandreOgeret:
Ok so my work here is done ? :)
Thanks for your help

nicolas-grekas:
If you can update the Travis matrix, that's still todo :)

MaxandreOgeret:
I guess, because it forces PHP7.1.
https://github.com/doctrine/common/issues/826
I think common will be removed in doctrine 3.0

Tobion:
Upping the php requirement does not require a new major version as it does not break bc.

nicolas-grekas:
It breaks the continuous migration path by forcing ppl to change two pieces of the infra at the same time (version of PHP + version of the package.)

Tobion:
They can update the php version without changing the SensioFrameworkExtraBundle. But this topic has been discussed alot of times. https://semver.org/#what-should-i-do-if-i-update-my-own-dependencies-without-changing-the-public-api

composer.json
```diff
@@ -15,7 +15,9 @@
         "symfony/framework-bundle": "^3.4|^4.0",
         "symfony/dependency-injection": "^3.3|^4.0",
         "symfony/http-kernel": "^3.3|^4.0",
-        "doctrine/common": "^2.2"
+        "doctrine/collections": "~1.0",
+        "doctrine/event-manager": "~1.0",
```
stof: does this bundle really use the Doctrine events ? AFAIK, it uses `doctrine/annotations` (missing here) and `doctrine/persistence`

composer.json
```diff
@@ -15,7 +15,9 @@
...
+        "doctrine/persistence": "~1.0"
```
stof: should use the caret operator for consistency.

nicolas-grekas:
@Tobion I'm not talking about semver but about the policy we have in Symfony. We have a pretty good track record over years of bumping PHP on major versions only. I don't see a reason to stop it now here.

Tobion:
I didn't know of this policy. At least in https://github.com/symfony/monolog-bundle/releases/tag/v3.3.0 it was violated then.

stof:
this one dropped support for EOLed PHP versions only. PHP 5.6 is not EOL yet

simensen:
Does this mean we can't merge this until EOL of PHP 5.6? Looks like that is at the end of the month? I'm getting warnings related to Doctrine's class loader in my message consumer and they are annoying since I cannot do anything about them. :(

Is there anything else that needs to be done on this PR to keep it from being merged on January 1st? :) Happy to see if I can pitch in on that if there is more to do.

stof:
this bundle is *not* using Doctrine ClassLoader. So this PR won't change that fact.

fabpot:
I've finished the work in #612

PR #585, opened 2018-09-03
Update the docs about Route and Method annotations

javiereguiluz:
This is still confusing for some users (see https://github.com/symfony/symfony-docs/issues/10142) so I propose to stop explaining the deprecated annotations and just explain how to upgrade.

### DIFF
--- a/Resources/doc/annotations/routing.rst
+++ b/Resources/doc/annotations/routing.rst
@@ -1,190 +1,61 @@
 @Route and @Method
 ==================
 
-.. caution::
+**Routing annotations of the SensioFrameworkExtraBundle are deprecated** since
+version 5.2 because they are now a core feature of Symfony.
 
-    Routing annotations of the SensioFrameworkExtraBundle are deprecated since version 5.2.
-    Configuring routes with annotations is now a core feature of Symfony. Please use it
-    instead as explained in :doc:`/routing`.
+How to Update your Applications
+-------------------------------
 
-Usage
------
+``@Route`` Annotation
+~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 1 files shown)


Resources/doc/annotations/routing.rst
```diff
@@ -1,190 +1,61 @@
...
     }
 
-The ``index`` action of the ``Post`` controller is now mapped to the ``/``
-URL. This is equivalent to the following YAML configuration:
+``@Method`` Annotation
+~~~~~~~~~~~~~~~~~~~~~~
 
-.. code-block:: yaml
+The ``@Method`` annotation from SensioFrameworkExtraBundle has been removed.
+Instead, the Symfony ``@Route`` annotation defines a new ``methods`` option to
```
xabbuh: The `methods` option was introduced in Symfony 2.2. So I would not call it "new" here.

Resources/doc/annotations/routing.rst
```diff
@@ -1,190 +1,61 @@
...
+How to Update your Applications
+-------------------------------
 
-Usage
------
+``@Route`` Annotation
+~~~~~~~~~~~~~~~~~~~~~
 
-The ``@Route`` annotation maps a route pattern with a controller::
+The Symfony annotation has the same options as the SensioFrameworkExtraBundle
```
xabbuh: The `service` option is not part of the core `@Route` annotation. Maybe it's worth to mention here that you should use FQCN as service ids now.

Resources/doc/annotations/routing.rst
xabbuh: see symfony/symfony#6738 for the PR adding it in Symfony

PR #586, opened 2018-09-07
remove reference to no longer existing section

xabbuh:
That section was removed in PR #585.


fabpot:
Good catch, thanks @xabbuh.

PR #594, opened 2018-11-30
Fix deprecation for symfony/config 4.2

### LINKED ISSUE #593: Symfony 4.2 deprecation: A tree builder without a root node is deprecated
`A tree builder without a root node is deprecated since Symfony 4.2 and will not be supported anymore in 5.0.`

It happens here:
https://github.com/sensiolabs/SensioFrameworkExtraBundle/blob/9ef408febe2f12e70118ef61c6515035a06c5830/DependencyInjection/Configuration.php#L32-L33

I suppose it should [be fixed like this](https://github.com/symfony/monolog-bundle/pull/281), right? To be fully backwards compatible. If so, I can prepare a PR for it.

mhujer:
Fixed the same way as in https://github.com/doctrine/DoctrineBundle/pull/853

Closes #593

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -29,8 +29,14 @@ class Configuration implements ConfigurationInterface
      */
     public function getConfigTreeBuilder()
     {
-        $treeBuilder = new TreeBuilder();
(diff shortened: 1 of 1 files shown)


mhujer:
Travis failure looks unrelated.

fabpot:
Thank you @mhujer.

PR #597, opened 2018-12-05
[DoctrineParamConverter][Docs] Add evict_cache documentation

IonBazan:
The documentation of `evict_cache` option was missing from `DoctrineParamConverter`. 
`map_method_signature` is undocumented too, but it is deprecated.


fabpot:
Thank you @IonBazan.

PR #600, opened 2018-12-11
Fix checking of invokable object controllers in ParamConverterListener

### LINKED ISSUE #599: Bug in checking controller callability ParamConverterListener
The second argument to `is_callable()` should be a `bool`. This line passes the string `__invoke`:

https://github.com/sensiolabs/SensioFrameworkExtraBundle/blob/1032c7077fd1a6f24f98b5a8377938000859f35d/EventListener/ParamConverterListener.php#L67

I suspect it's meant to be:

```php
} elseif (\is_object($controller) && \is_callable(array($controller, '__invoke'))) {
```

jaikdean:
This fixes #599


fabpot:
Thank you @jaikdean.

PR #602, opened 2019-01-15
Fix #574, split IsGrantedListener event

### LINKED ISSUE #574: Can't use IsGranted in combinason with the UserValueResolver
If you use the `UserValueResolver` to inject your user into a controller while using an `IsGranted` (or `Security`) annotation to limit the access to authenticated users like this

```php
/**
 * @IsGranted("ROLE_USER")
 */
public function myAction(UserInterface $user)
{
}
```

 you will get a 500 error for anonymous users because arguments resolving is done before triggering the listener that handle those annotations.

I'm not sure what is the best may solve this issue:
 - Putting the listener back to the controller event means we can't make decisions based on resolved arguments (unless we call the `ArgumentResolver` from the listener but that means running the resolving logic twice).
 - We could make the `UserValueResolver` throw a security exception if the user is anonymous, but it means the `IsGranted` annotation will still be ignored (and thus custom exception messages and status code wouldn't work).
 - We could make the argument resolving not throw immediately but instead wait after the `CONTROLLER_ARGUMENTS` event is dispatched to throw for missing arguments.

TomaszGasior:
Split IsGrantedListener::onKernelControllerArguments() event into
two separate events.

* New IsGrantedListener::onKernelController() is attached to
  KernelEvents::CONTROLLER and runs only @IsGranted annotations
  WITHOUT subject.
* IsGrantedListener::onKernelControllerArguments() is attached to
  KernelEvents::CONTROLLER_ARGUMENTS (no change here) but only
  runs @IsGranted annotations WITH subject.

This way simple @IsGranted annotations will be interpreted *before*
preparing arguments for controller action, which fixes #574.
As example — now, there is a way to type-hint UserInterface (from
Security bundle) in controller action without need to set nullable
type-hint.

### DIFF
--- a/EventListener/IsGrantedListener.php
+++ b/EventListener/IsGrantedListener.php
@@ -15,6 +15,7 @@ use Sensio\Bundle\FrameworkExtraBundle\Configuration\IsGranted;
 use Sensio\Bundle\FrameworkExtraBundle\Request\ArgumentNameConverter;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 use Symfony\Component\HttpKernel\Event\FilterControllerArgumentsEvent;
+use Symfony\Component\HttpKernel\Event\FilterControllerEvent;
 use Symfony\Component\HttpKernel\Exception\HttpException;
 use Symfony\Component\HttpKernel\KernelEvents;
 use Symfony\Component\Security\Core\Authorization\AuthorizationCheckerInterface;
@@ -36,6 +37,24 @@ class IsGrantedListener implements EventSubscriberInterface
         $this->authChecker = $authChecker;
     }
 
+    public function onKernelController(FilterControllerEvent $event)
+    {
+        $request = $event->getRequest();
+        /** @var $configurations IsGranted[] */
+        if (!$configurations = $request->attributes->get('_is_granted')) {
+            return;
+        }
+
+        $this->warnAboutAuthChecker();
+
+        foreach ($configurations as $configuration) {
+            // Only configurations WITHOUT subject.
+            if (null === $configuration->getSubject()) {
+                $this->checkIsGranted($configuration, null);
+            }
+        }
+    }
+
     public function onKernelControllerArguments(FilterControllerArgumentsEvent $event)
     {
         $request = $event->getRequest();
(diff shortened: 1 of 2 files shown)


TomaszGasior:
Unfortunately, my pull request fixes the problem only for one @IsGanted use case — when it's used WITHOUT subject. With subject specified, problem still exists and it's not possible to completely fix it with current Symfony architecture. But, for better developer experience, I added IsGrantedListener::onKernelException which warns Symfony user about unsupported usage of @IsGranted (with subject and UserInterface type hint).

https://i.imgur.com/jGmwfz9.png

It's a little bit hackish but in my opinion Symfony users are worth it. ;)

Koc:
I have the same problem, thanx for opening PR. Would be nice to merge it.
Only one thing that I don't understand - why are you subscribe to `kernel.exception` event?

// cc @fabpot

TomaszGasior:
>I have the same problem, thanx for opening PR. Would be nice to merge it.
>Only one thing that I don't understand - why are you subscribe to kernel.exception event?

My intention is to show for the user clear error message and instruct him how to fix the problem. Please see my comment: https://github.com/sensiolabs/SensioFrameworkExtraBundle/pull/602#issuecomment-455363970

mpoiriert:
When this pull request will be merge ?

Also a work around for using IsGranted with a subject is to have two is IsGranted annotation, one that validate a role (@IsGranted("ROLE_AUTHENTICATED")) and the other one with the subject. That way the first one will fail if the user is not authenticated and will stop everything else. Afterwards injection or annotation will be executed properly. This would be cleaner than have optional argument, will work with all the case. Still need this PR to be merge for it to work...

azjezz:
@TomaszGasior needs a rebase :)

nicolas-grekas:
@azjezz would you like to take over this PR? I'd then invite you to squash all commits from @TomaszGasior in one, then add yours + send a new PR, when you have some time :)

azjezz:
@nicolas-grekas sure :+1: :)

EventListener/IsGrantedListener.php
```diff
@@ -36,41 +41,92 @@ public function __construct(ArgumentNameConverter $argumentNameConverter, Author
...
+
+        foreach ($configurations as $configuration) {
+            // Only configurations WITHOUT subject.
+            if (null === $configuration->getSubject()) {
+                $this->checkIsGranted($configuration, null);
+            }
         }
+    }
+
+    public function onKernelControllerArguments(FilterControllerArgumentsEvent $event)
```
nicolas-grekas: this type hints a deprecated class, isn't it?
better remove the type hint and replace it with an annotation with the non-deprecated type
that'll preserve compat with 3.4 while preparing for 5.0
please check other event too

TomaszGasior:
Since I am busy on other work I hope @azjezz  will do good job. Sorry for late response. Closing this PR since @azjezz  may want create new. Thx!

PR #606, opened 2019-02-28
[Annotation] Add custom exceptions for IsGranted annotation

deguif:
I encountered recently a use case, where I need to intercept the exception thrown by the `@IsGranted` annotation, but currently I'm unable to identify such an exception, as the exceptions thrown are generic.

My proposal here is to use specific exceptions so that I can catch these one and know that they came from `@IsGranted`. I added the `attributes` and the `subject` used when the exception is thrown as they are valuable (at least for my use case) in this context.

Another solution could be to dispatch an event also.

### DIFF
--- a/EventListener/IsGrantedListener.php
+++ b/EventListener/IsGrantedListener.php
@@ -12,13 +12,13 @@
 namespace Sensio\Bundle\FrameworkExtraBundle\EventListener;
 
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\IsGranted;
+use Sensio\Bundle\FrameworkExtraBundle\Exception\IsGrantedAccessDeniedException;
+use Sensio\Bundle\FrameworkExtraBundle\Exception\IsGrantedHttpException;
 use Sensio\Bundle\FrameworkExtraBundle\Request\ArgumentNameConverter;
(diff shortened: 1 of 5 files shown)


deguif:
For more details about the use case.

I got a lot of controller with the same rules `@IsGranted`.
Instead of replacing the `@IsGranted` by a logic done inside the controller, I would like to catch Exception via a listener on `onKernelException` and do some logic based on the attribute/s and subject check that originally triggered the exception.

snebes:
You could use the current implementation of `@IsGranted` with a specific status code that you can catch.  i.e. 

`@IsGranted("ATTR", subject="subject", statusCode=499)`

```
    public function onKernelException(GetResponseForExceptionEvent $event)
    {
        $exception = $event->getException();

        if ($exception instanceof HttpException) {
             if (499 === $exception->getStatusCode() && ['xxx'] === $exception->getAttributes()) {
                 $event->setResponse(new RedirectResponse('xxxxxxxxx'));
             }
        }
```

deguif:
That's a solution, but I wouldn't get the original subject or attribute that triggered the exception ;)

deguif:
Nicer approach: https://github.com/sensiolabs/SensioFrameworkExtraBundle/pull/693
I'm closing this one.

PR #608, opened 2019-03-11
Wire nyholm/psr7 as default factory for PSR-7

nicolas-grekas:


DependencyInjection/SensioFrameworkExtraExtension.php
```diff
@@ -148,8 +152,13 @@ public function load(array $configs, ContainerBuilder $container)
         if ($config['psr_message']['enabled']) {
             $loader->load('psr7.xml');
 
-            if (!class_exists(ServerRequestFactory::class)) {
-                $definitionsToRemove[] = 'sensio_framework_extra.psr7.argument_value_resolver.server_request';
+            if (!interface_exists(StreamFactoryInterface::class)) {
+                throw new LogicException('PSR-7 support cannot be enabled as the required dependencies are not installed. Try running "composer require nyholm/psr7".');
+            }
+
+            if (class_exists(ServerRequestFactory::class) && !class_exists(Psr17Factory::class)) {
```
fabpot: I would remove this block

fabpot: And everything related to Diactorios to have only Nyholm/PSR-7 as a dev req.

nicolas-grekas: If we do that, we need to bump the minimum PHP version to 7.1. Should we do it?

fabpot: Let's do it (and let's bump the version to a new minor)

nicolas-grekas: Just while it was green :)
https://travis-ci.org/sensiolabs/SensioFrameworkExtraBundle/builds/504894762
pushed, let's see if it's still green with PHP 7.3 in

Resources/config/psr7.xml
```diff
@@ -5,7 +5,7 @@
            xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
 
     <services>
-        <service id="sensio_framework_extra.psr7.http_message_factory" class="Symfony\Bridge\PsrHttpMessage\Factory\DiactorosFactory" public="false" />
+        <service id="sensio_framework_extra.psr7.http_message_factory" class="Symfony\Bridge\PsrHttpMessage\Factory\PsrHttpFactory" public="false" autowire="true" />
```
nicolas-grekas: autowire="true", that's the trick to create decoupled configuration

nicolas-grekas: removed: explicit aliases to the rescue

fabpot:
Thank you @nicolas-grekas.

PR #609, opened 2019-03-25
[Docs] Remove action-suffix for contoller actions

chr-hertel:


fabpot:
Thank you @chr-hertel.

PR #618, opened 2019-05-24
Allow arrays of parameters to be used with @IsGranted

snebes:
This will allow `@IsGranted` annotation to pass an array of parameters to the security checker.  For example:

```
@IsGranted("CAN_DO_THINGS", subject={"thing1", "thing2"})
```

The equivalent can be done using an expression in the `@Security` annotation, but using the `@IsGranted` annotation provides better readability.


fabpot:
Thank you @snebes.

dmaicher:
So actually this should now be deprecated? :see_no_evil: 

See https://github.com/symfony/symfony/pull/33584

PR #624, opened 2019-07-02
Add example of `@IsGranted()` with multiple roles.

mattjanssen:
There are use cases of `@IsGranted()` where we want to check that a user has any one of multiple roles that don't intersect in the role hierarchy. I've seen this is two separate occasions already, especially with people migrating away from the JMSSecurityExtraBundle.

Does including one example make sense? Should there be a separate sentence or paragraph pointing this out?


fabpot:
Thank you @mattjanssen.

Issue #619, opened 2019-06-03
SF4.3 deprecations


drekinov:
Symfony\Component\Security\Core\Authentication\Token\AbstractToken::getRoles()
Symfony\Component\Security\Core\Role\RoleHierarchy::getReachableRoles()
are deprecated. Both are called from EventListener\SecurityListener.php:90

PR #625, opened 2019-07-03
Fixed Symfony 4.3 deprecations.

tarlepp:
Solves #619


fabpot:
Thank you @tarlepp.

tarlepp:
@fabpot I think that you merged this a bit too early, with low deps tests will fail. Need to think something to solve that.

PR #626, opened 2019-07-03
Fixed behaviour on Symfony 3.4

tarlepp:
Not 100% sure if  this is the "proper" way to handle those event class changes properly - but at least now tests are ok with 3.4 and 4.3 - although I don't like to use that generic `KernelEvent` type on those listener classes - but didn't found another solution to fix that.

If there is another solution for that I'm like to hear that - so that I could improve this PR.

Related to #625


fabpot:
Thank you @tarlepp.

tarlepp:
Hmm, I see that construct signature is different with that one - As I said in my initial comment that I didn't like this solution - although all the tests are green with SF3.4 and 4.3.

And that is why I asked if there was another solution to fix this - or is this something that cannot be fixed before SF5 to get support for SF3.4 and 4.3?

I'm open for any ideas you might have.

PR #627, opened 2019-07-05
Load Tests with autoloader only in dev

rvitaliy:
Hi, this PR allow us to exclude Tests from production autoloading


fabpot:
Thank you @rvitaliy.

PR #629, opened 2019-07-05
Fix has_role compatibility in Symfony 3.4

### LINKED ISSUE #628: has_role is broken in Symfony 3.4 in 5.4.0
Because of this change https://github.com/sensiolabs/SensioFrameworkExtraBundle/commit/e5f9bbf26fabcc1e30ef75c76e6de689a063c5d9#diff-9bba23159f3d837b5f7261925414651fL101

This line https://github.com/symfony/symfony/blob/d9833ace32c69681a0a01f309f5f2b6cc57b0485/src/Symfony/Component/Security/Core/Authorization/ExpressionLanguageProvider.php#L54 now receives array of objects instead of array of strings and fails. 

We have hundreds of tests failing with HTTP 403 error

ostrolucky:
Fixes #628 

Looked into tests too, but didn't see any simple way to write a test for 3.4 only, while avoiding deprecations

### DIFF
--- a/EventListener/SecurityListener.php
+++ b/EventListener/SecurityListener.php
@@ -98,6 +98,10 @@ class SecurityListener implements EventSubscriberInterface
             } else {
                 $roles = $token->getRoles();
             }
+
+            $roles = array_map(function ($role) {
+                return $role->getRole();
+            }, $roles);
         }
 
         $variables = [


plozmun:
This solve our 403 issues too. Thank you @ostrolucky

jtaylor100:
Ours too. Thank you @ostrolucky

fabpot:
Thank you @ostrolucky.

PR #630, opened 2019-07-19
[@IsGranted] Updated exception handling when the $subject is an array of values

snebes:
Adding better exception handling functionality for the subject array capability added in #618.  This will cover the use cases of a single element array, or multiple element array.


fabpot:
See #783

PR #631, opened 2019-07-24
Fix #574, split IsGrantedListener event

### LINKED ISSUE #574: Can't use IsGranted in combinason with the UserValueResolver

azjezz:
Split IsGrantedListener::onKernelControllerArguments() event into
two separate events.

* New IsGrantedListener::onKernelController() is attached to
  KernelEvents::CONTROLLER and runs only @IsGranted annotations
  WITHOUT subject.
* IsGrantedListener::onKernelControllerArguments() is attached to
  KernelEvents::CONTROLLER_ARGUMENTS (no change here) but only
  runs @IsGranted annotations WITH subject.

This way simple @IsGranted annotations will be interpreted *before*
preparing arguments for controller action, which fixes #574.
As example — now, there is a way to type-hint UserInterface (from
Security bundle) in controller action without need to set nullable
type-hint.


EventListener/IsGrantedListener.php
```diff
@@ -36,54 +41,95 @@ public function __construct(ArgumentNameConverter $argumentNameConverter, Author
         $this->authChecker = $authChecker;
     }
 
-    public function onKernelControllerArguments(KernelEvent $event)
+    public function onKernelController(FilterControllerEvent $event)
```
nicolas-grekas: same comment as on the other method, this type is deprecated, please check them all

EventListener/IsGrantedListener.php
```diff
@@ -36,54 +41,95 @@ public function __construct(ArgumentNameConverter $argumentNameConverter, Author
...
+        foreach ($configurations as $configuration) {
+            // Only configurations WITHOUT subject.
+            if (null === $configuration->getSubject()) {
+                $this->checkIsGranted($configuration, null);
+            }
+        }
+    }
+
+    /**
+     * @param \Symfony\Component\HttpKernel\Event\ControllerArgumentsEvent|FilterControllerArgumentsEvent $event
```
nicolas-grekas: we should use a `use` statement + we document only the non deprecated way

ottaviano:
hi @azjezz, thank you for this job, do you think you have time to finish the PR?

fabpot:
See #783

PR #633, opened 2019-08-05
Improve DateTimeParamConverter documentation

enumag:
With `Y-m-d` format you don't care about the time. In this case I think it's better to force midnight instead of the current time.


fabpot:
Nice, is it something documented somewhere?

andrew-demb:
Yes

```
! | Resets all fields (year, month, day, hour, minute, second, fraction and timezone information) to the Unix Epoch | Without !, all fields will be set to the current date and time.
```
https://www.php.net/manual/en/datetime.createfromformat.php

fabpot:
Thank you @enumag.

PR #635, opened 2019-08-12
use Symfony\Component\Routing\Annotation\Route;

LaurentMarquet:
Shouldn't we read `use Symfony\Component\Routing\Annotation\Route;`?


fabpot:
Thank you @LaurentMarquet.

PR #640, opened 2019-09-30
rm dependency on doctrine/persistence

dmaicher:
See https://github.com/symfony/symfony/issues/33164

- make `doctrine/persistence` optional since it will be required transitively anyway via the ORM/ODM that is required for using `DoctrineParamConverter`


fabpot:
Thank you @dmaicher.

PR #643, opened 2019-10-03
Ensure owner of Template is not null

ostrolucky:
Some libraries trust definition of getOwner method, but it will return null if no owner was set. My personal experience with this was Rector which crashes, as it passses output of this method to function requiring array. Rather than contributing cast patch to rector, I think it would be better to add the missing default directly here.


fabpot:
Thank you @ostrolucky.

PR #647, opened 2019-10-15
Fix #646 : Add test exists Doctrine\Common\Persistence\Proxy

### LINKED ISSUE #646: Missing doctrine/persistence
Hi,

They is no more doctrine/persistence with dev-master but still use it in TemplateGuesser and ControllerListener.

mazarini:
Suppress error when doctrine/persistence is not require


fabpot:
Thank you @mazarini.

PR #648, opened 2019-10-22
Deprecated passing more than one attribute to @IsGranted

### LINKED ISSUE #645: SF 4.4/5.0 deprecates/removes IsGranted multi-attribute usage.
https://github.com/symfony/symfony/pull/33584 removes the ability to pass an array of attributes to `isGranted()` and `decide()` methods. Should the `@IsGranted()` annotation be updated to not allow arrays, too?

If so, let's not forget to revert this documentation commit (https://github.com/sensiolabs/SensioFrameworkExtraBundle/commit/97cc360add63a1376bbbe8e1b912c43c07640ef2) I made in July.

wouterj:
Fixes #645 

I'm not so sure about the version, but looking at the tags, I expected the next version will be 5.6 (as 5.5 is already released?).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,12 @@
 CHANGELOG
 =========
 
+5.6
+---
+
+ * Deprecated passing more than one attribute to `@IsGranted()`. Use an expression instead
+   (e.g. `@IsGranted("has_role('ROLE_USER') or has_role('ROLE_ADMIN')")`).
+
 5.2
 ---
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ CHANGELOG
 ---
 
  * Deprecated passing more than one attribute to `@IsGranted()`. Use an expression instead
-   (e.g. `@IsGranted("has_role('ROLE_USER') or has_role('ROLE_ADMIN')")`).
+   (e.g. `@Security("has_role('ROLE_USER') or has_role('ROLE_ADMIN')")`).
 
 5.2
 ---
--- a/src/EventListener/IsGrantedListener.php
+++ b/src/EventListener/IsGrantedListener.php
@@ -73,7 +73,7 @@ class IsGrantedListener implements EventSubscriberInterface
                 }
             }
 
-            if (!$this->authChecker->isGranted($configuration->getAttributes(), $subject)) {
+            if (!$this->authChecker->isGranted($configuration->getAttribute(), $subject)) {
                 $argsString = $this->getIsGrantedString($configuration);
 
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,12 @@
 CHANGELOG
 =========
 
+5.6
+---
+
+ * Deprecated passing more than one attribute to `@IsGranted()`. Use an expression instead
+   (e.g. `@IsGranted("has_role('ROLE_USER') or has_role('ROLE_ADMIN')")`).
```
jvasseur:
```suggestion
   (e.g. `@Security("has_role('ROLE_USER') or has_role('ROLE_ADMIN')")`).
```
`IsGranted` doesn't work with an expression (because you would need to pass an `Expression` object).

wouterj:
Thanks @jvasseur! I'm not using this bundle, so wasn't aware of the difference between those annotations.

CHANGELOG.md
```diff
@@ -1,6 +1,12 @@
 CHANGELOG
 =========
 
+5.6
+---
+
+ * Deprecated passing more than one attribute to `@IsGranted()`. Use an expression instead
+   (e.g. `@Security("has_role('ROLE_USER') or has_role('ROLE_ADMIN')")`).
```
weaverryan:
```suggestion
   (e.g. `@Security("is_granted('ROLE_USER') or is_granted('ROLE_ADMIN')")`).
```

I think `is_granted()` is a more accurate replacement

src/EventListener/IsGrantedListener.php
```diff
@@ -73,7 +73,7 @@ public function onKernelControllerArguments(KernelEvent $event)
                 }
             }
 
-            if (!$this->authChecker->isGranted($configuration->getAttributes(), $subject)) {
+            if (!$this->authChecker->isGranted($configuration->getAttribute(), $subject)) {
```
weaverryan: If you're passing an array of attributes and are using Symfony 5, you will see the deprecation from this bundle, but the feature won't work correctly because Symfony 5 doesn't support an array of attributes, right?

I think we need to either:

A) "Mimic" the "or" behavior here - i.e. if it's an array, we call `isGranted()` twice in an "or" kind of way
B) Throw an actual exception if you're using Symfony 5 with this bundle and are passing an array of attributes. That's a BC Break, but an obvious one (better than the silent one of it just not working) AND the user *should* have seen deprecation warnings on Symfony 4.4 anyways. I think this would prevent WTF's like this one: https://github.com/sensiolabs/SensioFrameworkExtraBundle/issues/645#issuecomment-569914831

fabpot:
See #783

PR #650, opened 2019-11-06
Add cache-control `stale-while-revalidate` and `stale-if-error` support

franek:
These headers can be used by proxy cache (Varnish, Nginx) in order to serve stale contents.


fabpot:
Thank you @franek.

PR #655, opened 2019-12-19
Update converters.rst

Naxirdan:
Piece of code were we can see how we could implement evict_cache param


fabpot:
I don't think we need an example here. There are already similar examples before.

PR #656, opened 2019-12-26
don't fail when token storage is an alias

### LINKED ISSUE #653: @Security annotation evaluation not working as expected
Hello there!

After upgrading to Symfony 4.4 (in particular 4.4.1) from Symfony 4.3.8 the security annotation expressions in my controllers stopped working as expected. I saw the deprecation notices about using an array, but that's not my use case.

What happened:
```php
<?php declare(strict_types=1);

namespace App\Controller;

use Sensio\Bundle\FrameworkExtraBundle\Configuration\IsGranted;
use Sensio\Bundle\FrameworkExtraBundle\Configuration\Security;
use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
use Symfony\Component\HttpFoundation\Response;
use Symfony\Component\Routing\Annotation\Route;

class MyController extends AbstractController
{
    /**
     * @Route(path="/items", methods={"POST"})
     *
     * @Security("is_granted('ROLE_ADMIN')")
     * @IsGranted("ROLE_ADMIN")
     */
    public function postItemAction(Request $request): Response
    {
       // code...
    }
}
```

If the `@IsGranted` annotation is used the `ROLE_ADMIN` is checked with the `RoleHierarchyVoter` and everything is fine. If the `@Security` annotation is used, the expression is not checked.

Am I missing something?

Thanks for your help 😄

xabbuh:
fixes #653

### DIFF
--- a/src/DependencyInjection/Compiler/OptimizerPass.php
+++ b/src/DependencyInjection/Compiler/OptimizerPass.php
@@ -23,7 +23,7 @@ class OptimizerPass implements CompilerPassInterface
 {
     public function process(ContainerBuilder $container)
     {
-        if (!$container->hasDefinition('security.token_storage')) {
+        if (!$container->has('security.token_storage')) {
             $container->removeDefinition('sensio_framework_extra.security.listener');
         }
 


fabpot:
Thank you @xabbuh.

PR #663, opened 2019-12-27
Use new Proxy class when possible

odolbeau:
`Doctrine\Common\Persistence\Proxy` have been deprecated in favor of `Doctrine\Persistence\Proxy` since `doctrine/persistence` 1.3

This PR deal with both Proxy class to avoid deprecation notice when the new Proxy class is available.


fabpot:
Thank you @odolbeau.

PR #666, opened 2020-01-09
Add Expression Language installation suggestion

nebkam:
In newer versions of Symfony (I've noticed it on 4.4 but it may appeared earlier), @Security annotation has a dependency on Expression Language Component.

### DIFF
--- a/src/Resources/doc/annotations/security.rst
+++ b/src/Resources/doc/annotations/security.rst
@@ -77,6 +77,12 @@ allows you to pass an *expression* that can contain custom logic::
         // ...
     }
 
+To use the ``@Security`` annotation, you need to have Expression Language Component installed.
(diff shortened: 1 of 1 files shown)


fabpot:
IIRC, the error message is explicit about installing it, right? If that's the case, there is no need to add it in the docs IMHO.

nebkam:
Yes, I agree the exception message is explicit:
`To use the @Security tag, you need to use the Security component 2.4 or newer and install the ExpressionLanguage component.`
The reason I've submitted this PR is because I thought it would make DX better to state the dependency upfront, without the need to run the code and fail, only to find out it was a hidden dependency.

fabpot:
It's not hidden, just optional. I still think that we do not need something in the docs. Shorter docs are always better.

PR #669, opened 2020-02-24
Update index.rst

abdounikarim:
Hello, i juste removed an alias from the docs. 😉


fabpot:
Annotations are required.

PR #673, opened 2020-03-12
remove irrelevant code from unsupported php versions

Tobion:
the bundle requires php 7.1 at the moment


fabpot:
Thank you @Tobion.

PR #677, opened 2020-04-09
Fix some namespace issues in the tests directory

iamdto:
We noticed these warnings when running the following command in our projects:

```
$ composer install  --optimize-autoloader
Loading composer repositories with package information
Installing dependencies (including require-dev) from lock file
Nothing to install or update
Generating optimized autoload files
Deprecation Notice: Class Sensio\Bundle\FrameworkExtraBundle\Tests\Request\ParamConverter\ArgumentNameConverterTest located in ./vendor/sensio/framework-extra-bundle/Tests/Request/ArgumentNameConverterTest.php does not comply with psr-4 autoloading standard. It will not autoload anymore in Composer v2.0. in phar:///usr/local/Cellar/composer/1.10.1/bin/composer/src/Composer/Autoload/ClassMapGenerator.php:201
Deprecation Notice: Class Sensio\Bundle\FrameworkExtraBundle\Tests\DependencyInjection\AddParamConverterPassTest located in ./vendor/sensio/framework-extra-bundle/Tests/DependencyInjection/Compiler/AddParamConverterPassTest.php does not comply with psr-4 autoloading standard. It will not autoload anymore in Composer v2.0. in phar:///usr/local/Cellar/composer/1.10.1/bin/composer/src/Composer/Autoload/ClassMapGenerator.php:201
Deprecation Notice: Class Sensio\Bundle\FrameworkExtraBundle\Tests\DependencyInjection\AddExpressionLanguageProvidersPassTest located in ./vendor/sensio/framework-extra-bundle/Tests/DependencyInjection/Compiler/AddExpressionLanguageProvidersPassTest.php does not comply with psr-4 autoloading standard. It will not autoload anymore in Composer v2.0. in phar:///usr/local/Cellar/composer/1.10.1/bin/composer/src/Composer/Autoload/ClassMapGenerator.php:201
```

If you try to run the same command in this repository you will not be able to see them because of the following lines:

https://github.com/sensiolabs/SensioFrameworkExtraBundle/blob/d0585d4825a87a5030ca8cd34adb4a17e1066c17/composer.json#L46-L48

This PR should fix the warnings, however I'm wondering if that configuration in the `composer.json` file is necessary at all? Isn't it better to install with `--no-dev` if you want to exclude test classes from your autoloader? That configuration currently "hides" these warnings (if it'd have to happen again in the future).


iamdto:
On second thought I think removing the `exclude-from-classmap` would only partially solve the problem. Most developers probably just run `composer install` if they want to contribute to this project, so even with this removed you would still not see the warnings.

fabpot:
Thank you @iamdto.

PR #680, opened 2020-04-22
Add throw_notfound option to ParamConverter

ph-fritsche:
### New option
The proposed changes add a new boolean option `throw_notfound`.
When set to true, the `DoctrineParamConverter` throws a `NotFoundHttpException` for optional parameters if identifier for find() or search criteria for findOneBy() are given and no corresponding object can be found or value conversion by Doctrine fails.

```php
/**
 * @Route("/foo/bar/{id}")
 * @ParamConverter("baz", options={"throw_notfound": true})
 */
public function __invoke(?Baz $baz) {
```

```php
/**
 * @Route("/foo/{bar}/{baz}")
 * @ParamConverter("foo", options={"throw_notfound": true})
 */
public function __invoke(?Foo $foo) {
```

### Prevent running findOneBy() for identifier

When an identifier parameter was set to null, `DoctrineParamConverter::apply()` triggered a database query with `WHERE id IS NULL` per `findOneBy()`.
The proposed changes prevent null value for identifiers to fall through to `DoctrineParamConverter::findOneBy()`.

### DIFF
--- a/src/Request/ParamConverter/DoctrineParamConverter.php
+++ b/src/Request/ParamConverter/DoctrineParamConverter.php
@@ -88,14 +88,14 @@ class DoctrineParamConverter implements ParamConverterInterface
                 $object = $this->findViaExpression($class, $request, $expr, $options, $configuration);
 
                 if (null === $object) {
-                    $errorMessage = sprintf('The expression "%s" returned null', $expr);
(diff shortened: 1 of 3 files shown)


fabpot:
See #783

PR #681, opened 2020-05-06
Fetch Twig lazily in TemplateListener

nicolas-grekas:
I spotted this when playing with a controller that doesn't use Twig:

`TemplateListener` always instantiate twig. But twig is kinda heavy to create. This makes its instantiation lazy.


fabpot:
Thank you @nicolas-grekas.

PR #686, opened 2020-06-10
Add `@ContentSafe` annotation for controller actions

rvanlaak:
In Symfony 5.2 support was added for safe content negotiation via PR https://github.com/symfony/symfony/pull/34847 and was explained via https://symfony.com/blog/new-in-symfony-5-1-the-safe-http-preference

This PR introduces the `@ContentSafe` annotation. The annotation automatically will get enabled when `Response::setContentSafe` method exists, or by setting `content_safe.annotations` to `true` in the configuration.

## Controller actions do not necessarily create the `Response` object themselves. 

Annotations like `@Template` can get used to do that.

```php
/*
 * @Template()
 */
public function indexAction()
{
    return ['data' => $data];
}
```

Without the `@ContentSafe` annotation the `@Template` annotation can not get used anymore. With the `@ContentSafe` annotation this will look like:
```php
/*
 * @Template()
 * @ContentSafe
 */
public function indexAction()
{
    return ['data' => $data];
}
```

## Marking a `Response` with `setContentSafe` requires defining the response as variable.

Without content safe:
```php
return new Response(200, '<html><body>Hello world!</body></html>');
```
With content safe:
```php
$response = new Response(200, '<html><body>Hello world!</body></html>');
$response->setContentSafe();
return $response;
```


rvanlaak:
@fabpot what do you think of this PR? Would be great to have the `@ContentSafe` annotation at our disposal.

fabpot:
See #783

PR #688, opened 2020-06-16
Add missing wiring for HttpFoundationFactoryInterface and HttpMessageFactoryInterface

drupol:
Hi,

In our project, we are heavily relying on PSR7 request and responses.

In order to get it working, we need to add those two lines for proper wiring, but I don't think they belong in our project, they should be provided by this bundle.


fabpot:
Thank you @drupol.

drupol:
Merci :-)

PR #693, opened 2020-07-01
Set attributes and subject to AccessDeniedException

### LINKED ISSUE #692: Throwing AccessDeniedException without setting attributes
When I use `$this->denyAccessUnlessGranted('SOME_ROLE');` in a controller, it triggers the following code:

```
        if (!$this->isGranted($attributes, $subject)) {
            $exception = $this->createAccessDeniedException($message);
            $exception->setAttributes($attributes);
            $exception->setSubject($subject);

            throw $exception;
        }
```

so the exception has the info about the attribute that caused it.

But when I replace `denyAccessUnlessGranted` with annotations from this bundle (`@IsGranted("SOME_ROLE")`, or `Security("is_granted('SOME_ROLE')")`), it simply throws an `AccessDeniedException` without populating it with attributes and subject.

```
throw new AccessDeniedException($message);
```

It's a useful feature that allows knowing which attribute caused the exception and it would be great to have it in this bundle too.

77web:
fix #692 

I've confirmed both `AccessDeniedException::$attributes` and `AccessDeniedException::$subject` exist in Symfony4.4+.
This Bundle restricts Symfony version to 4.4 or 5.1, so I think this PullRequest can be merged to master.

### DIFF
--- a/src/EventListener/IsGrantedListener.php
+++ b/src/EventListener/IsGrantedListener.php
@@ -82,7 +82,11 @@ class IsGrantedListener implements EventSubscriberInterface
                     throw new HttpException($statusCode, $message);
                 }
 
-                throw new AccessDeniedException($message);
+                $accessDeniedException = new AccessDeniedException($message);
+                $accessDeniedException->setAttributes($configuration->getAttributes());
+                $accessDeniedException->setSubject($subject);
+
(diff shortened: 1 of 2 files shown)


fabpot:
Thank you @77web.

PR #696, opened 2020-08-12
Add consistency by always throwing an HttpException

VincentLanglet:
Before, the listener was mixing 
```
Symfony\Component\Security\Core\Exception\AccessDeniedException
```
with 
```
use Symfony\Component\HttpKernel\Exception\HttpException;
```

I think we should use 
```
use Symfony\Component\HttpKernel\Exception\AccessDeniedHttpException;
```
instead because it extends HttpException.


VincentLanglet:
My bad, I understand the difference.

PR #703, opened 2020-11-18
Allow enabling null subject for IsGranted

duffpl:
**What is changed with this PR?**
Enable allowing null subject in IsGranted annotation.

**Why this feature was created?**
In one of my projects I have nullable action argument and voter that supports nullable subject. Unfortunately without allowing null subject I can't use IsGranted annotation.


fabpot:
See #783

PR #707, opened 2020-12-08
Add PHP 8.0 attributes support

Yoann-TYT:
# Add PHP 8.0 attributes support

## Using PHP attributes if we start a new project on PHP 8.0

Instead of using classic doctrine annotations 
```php
    /**
     * @Cache(public=true, maxage="0", smaxage="1 month")
     * @IsGranted("ROLE_ADMIN")
     */
    public function foo() { }
```

We should use attributes PHP 8.0 
```php
    #[Cache(public: true, maxage: '0', smaxage: '1 month')]
    #[IsGranted('ROLE_ADMIN')]
    public function foo() { }
```

### Note for Etag attribute

In the documentation, we use Etag ( not etag or ETag used in the unit tests ). So, to simplify the code, I suggest to fix the unit test. What do you think about that ? 

Thank you for your reading :)

### DIFF
--- a/src/Configuration/Cache.php
+++ b/src/Configuration/Cache.php
@@ -17,6 +17,7 @@ namespace Sensio\Bundle\FrameworkExtraBundle\Configuration;
  * @author Fabien Potencier <fabien@symfony.com>
  * @Annotation
  */
+#[\Attribute(\Attribute::TARGET_CLASS | \Attribute::TARGET_METHOD)]
 class Cache extends ConfigurationAnnotation
 {
     /**
@@ -30,7 +31,7 @@ class Cache extends ConfigurationAnnotation
      * The number of seconds that the response is considered fresh by a private
      * cache like a web browser.
      *
-     * @var int
+     * @var int|string|null
      */
     private $maxage;
 
@@ -38,7 +39,7 @@ class Cache extends ConfigurationAnnotation
      * The number of seconds that the response is considered fresh by a public
      * cache like a reverse proxy cache.
      *
-     * @var int
+     * @var int|string|null
      */
     private $smaxage;
 
@@ -100,6 +101,42 @@ class Cache extends ConfigurationAnnotation
      * @var int|string
      */
     private $staleIfError;
+
+    /**
+     * @param int|string|null $maxage
+     * @param int|string|null $smaxage
+     * @param int|string|null $maxstale
+     * @param int|string|null $staleWhileRevalidate
+     * @param int|string|null $staleIfError
+     */
+    public function __construct(
+        array $values = [],
+        string $expires = null,
+        $maxage = null,
+        $smaxage = null,
+        bool $public = false,
+        bool $mustRevalidate = false,
+        array $vary = [],
+        ?string $lastModified = null,
+        ?string $etag = null,
+        $maxstale = null,
+        $staleWhileRevalidate = null,
+        $staleIfError = null
+    ) {
+        $values['expires'] = $values['expires'] ?? $expires;
+        $values['maxage'] = $values['maxage'] ?? $maxage;
+        $values['smaxage'] = $values['smaxage'] ?? $smaxage;
+        $values['public'] = $values['public'] ?? $public;
+        $values['mustRevalidate'] = $values['mustRevalidate'] ?? $mustRevalidate;
+        $values['vary'] = $values['vary'] ?? $vary;
+        $values['lastModified'] = $values['lastModified'] ?? $lastModified;
+        $values['Etag'] = $values['Etag'] ?? $etag;
+        $values['maxstale'] = $values['maxstale'] ?? $maxstale;
+        $values['staleWhileRevalidate'] = $values['staleWhileRevalidate'] ?? $staleWhileRevalidate;
+        $values['staleIfError'] = $values['staleIfError'] ?? $staleIfError;
+
+        parent::__construct($values);
+    }
 
     /**
      * Returns the expiration date for the Expires header field.
--- a/src/Configuration/Entity.php
+++ b/src/Configuration/Entity.php
@@ -17,6 +17,7 @@ namespace Sensio\Bundle\FrameworkExtraBundle\Configuration;
  * @author Ryan Weaver <ryan@knpuniversity.com>
  * @Annotation
  */
+#[\Attribute(\Attribute::IS_REPEATABLE | \Attribute::TARGET_CLASS | \Attribute::TARGET_METHOD)]
 class Entity extends ParamConverter
 {
     public function setExpr($expr)
@@ -25,5 +26,30 @@ class Entity extends ParamConverter
         $options['expr'] = $expr;
 
         $this->setOptions($options);
+    }
+
+    /**
+     * @param array|string $data
+     */
+    public function __construct(
+        $data = [],
+        string $expr = null,
+        string $class = null,
+        array $options = [],
+        bool $isOptional = false,
+        string $converter = null
+    ) {
+        $values = [];
+        if (\is_string($data)) {
+            $values['value'] = $data;
+        } else {
+            $values = $data;
+        }
+
+        $values['expr'] = $values['expr'] ?? $expr;
+
+        parent::__construct($values, $class, $options, $isOptional, $converter);
+
+        $this->setExpr($values['expr']);
     }
 }
--- a/src/Configuration/IsGranted.php
+++ b/src/Configuration/IsGranted.php
@@ -17,6 +17,7 @@ namespace Sensio\Bundle\FrameworkExtraBundle\Configuration;
  * @author Ryan Weaver <ryan@knpuniversity.com>
  * @Annotation
  */
+#[\Attribute(\Attribute::IS_REPEATABLE | \Attribute::TARGET_CLASS | \Attribute::TARGET_METHOD)]
 class IsGranted extends ConfigurationAnnotation
 {
     /**
@@ -49,6 +50,29 @@ class IsGranted extends ConfigurationAnnotation
      * @var int|null
      */
     private $statusCode;
+
+    /**
+     * @param mixed        $subject
+     * @param array|string $data
+     */
+    public function __construct(
+        $data = [],
+        $subject = null,
+        string $message = null,
+        ?int $statusCode = null
+    ) {
+        $values = [];
+        if (\is_string($data)) {
+            $values['attributes'] = $data;
+        } else {
+            $values = $data;
+        }
+
+        $values['subject'] = $values['subject'] ?? $subject;
+        $values['message'] = $values['message'] ?? $message;
+        $values['statusCode'] = $values['statusCode'] ?? $statusCode;
+        parent::__construct($values);
+    }
 
     public function setAttributes($attributes)
     {
--- a/src/Configuration/ParamConverter.php
+++ b/src/Configuration/ParamConverter.php
@@ -17,6 +17,7 @@ namespace Sensio\Bundle\FrameworkExtraBundle\Configuration;
  * @author Fabien Potencier <fabien@symfony.com>
  * @Annotation
  */
+#[\Attribute(\Attribute::IS_REPEATABLE | \Attribute::TARGET_CLASS | \Attribute::TARGET_METHOD)]
 class ParamConverter extends ConfigurationAnnotation
 {
     /**
@@ -53,6 +54,29 @@ class ParamConverter extends ConfigurationAnnotation
      * @var string
      */
     private $converter;
+
+    /**
+     * @param array|string $data
+     */
+    public function __construct(
+        $data = [],
+        string $class = null,
+        array $options = [],
+        bool $isOptional = false,
+        string $converter = null
+    ) {
+        $values = [];
+        if (\is_string($data)) {
+            $values['value'] = $data;
+        } else {
+            $values = $data;
+        }
+        $values['class'] = $values['class'] ?? $class;
+        $values['options'] = $values['options'] ?? $options;
+        $values['isOptional'] = $values['isOptional'] ?? $isOptional;
+        $values['converter'] = $values['converter'] ?? $converter;
+        parent::__construct($values);
+    }
 
     /**
      * Returns the parameter name.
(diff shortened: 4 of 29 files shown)


src/EventListener/ControllerListener.php
```diff
@@ -60,6 +60,22 @@ public function onKernelController(KernelEvent $event)
...
+            array_walk($classAttributes, function (\ReflectionAttribute &$value) {
+                $className = '\\'.$value->getName();
+                $value = new $className($value->getArguments());
+            });
+            $classConfigurations = array_merge($classConfigurations, $this->getConfigurations($classAttributes));
+
+            $methodAttributes = $method->getAttributes();
+            array_walk($methodAttributes, function (\ReflectionAttribute &$value) {
+                $className = '\\'.$value->getName();
+                $value = new $className($value->getArguments());
```
maxhelias, line 74: A ReflectionAttribute has a `newInstance` property which does the work for us

composer.json
```diff
@@ -11,7 +11,7 @@
         }
     ],
     "require": {
-        "php": ">=7.2.5",
+        "php": ">=7.2.5|^8.0",
```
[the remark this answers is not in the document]
Yoann-TYT: Indeed :) Thanks !

src/Configuration/Cache.php
```diff
@@ -17,6 +17,7 @@
  * @author Fabien Potencier <fabien@symfony.com>
  * @Annotation
  */
+#[\Attribute()]
```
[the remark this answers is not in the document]
Yoann-TYT: Yes, I set up the targets IS_REPEATABLE, TARGET_CLASS et TARGET_CLASS :)

src/EventListener/ControllerListener.php
Yoann-TYT: it's so much easier ;)
Thanks for your review @maxhelias !

pkly:
Seems good to me

fabpot:
@derrabus Do you have time to review this PR?

src/EventListener/ControllerListener.php
```diff
@@ -60,6 +60,20 @@ public function onKernelController(KernelEvent $event)
         $classConfigurations = $this->getConfigurations($this->reader->getClassAnnotations($object));
         $methodConfigurations = $this->getConfigurations($this->reader->getMethodAnnotations($method));
 
+        if (method_exists($method, 'getAttributes')) {
```
derrabus, line 63:
```suggestion
        if (80000 >= \PHP_VERSION_ID) {
```

OPcache can optimize PHP version conditions and we can find this piece easily if we build a release without PHP 7 support one day.

src/EventListener/ControllerListener.php
```diff
@@ -60,6 +60,20 @@ public function onKernelController(KernelEvent $event)
...
+            $classAttributes = $object->getAttributes();
+            array_walk($classAttributes, function (\ReflectionAttribute &$value) {
+                $value = $value->newInstance();
+            });
```
derrabus, line 67: You are creating new instances of _every_ attribute that is attached to the class. This is something you should never do. The class might have all kinds of attributes, not only those that come from this bundle. Your code will break if there are attributes attached that cannot be instantiated, e.g. if the referenced attribute class is not defined.

Please only instantiate attributes that are relevant for you!

src/EventListener/ControllerListener.php
```diff
@@ -60,6 +60,20 @@ public function onKernelController(KernelEvent $event)
...
+            $classConfigurations = array_merge($classConfigurations, $this->getConfigurations($classAttributes));
+
+            $methodAttributes = $method->getAttributes();
+            array_walk($methodAttributes, function (\ReflectionAttribute &$value) {
+                $value = $value->newInstance();
+            });
```
derrabus, line 73: Can we use `array_map()` instead?

src/Configuration/Cache.php
```diff
@@ -17,6 +17,7 @@
  * @author Fabien Potencier <fabien@symfony.com>
  * @Annotation
  */
+#[\Attribute(\Attribute::IS_REPEATABLE | \Attribute::TARGET_CLASS | \Attribute::TARGET_METHOD)]
```
derrabus, line 20: Does it make sense to have multiple `Cache` attributes on the same class/method?

src/Configuration/Cache.php
```diff
@@ -101,6 +102,35 @@ class Cache extends ConfigurationAnnotation
...
+        $smaxage = null,
+        $public = false,
+        $mustRevalidate = false,
+        $vary = [],
+        $lastModified = null,
+        $Etag = null,
+        $maxstale = null,
+        $staleWhileRevalidate = null,
+        $staleIfError = null
+    ) {
```
derrabus, line 118: Please add parameter types where possible on PHP 7.2. Please add a `@param` annotation if you cannot define a property type.

Same with other constructors in this PR.

src/Configuration/Cache.php
```diff
@@ -101,6 +102,35 @@ class Cache extends ConfigurationAnnotation
...
+    public function __construct(
+        $values = [],
+        $expires = null,
+        $maxage = null,
+        $smaxage = null,
+        $public = false,
+        $mustRevalidate = false,
+        $vary = [],
+        $lastModified = null,
+        $Etag = null,
```
derrabus, line 114: The capital `E` looks odd.

derrabus (changes requested):
Thank you for working on this topic! I have added a few comments.

If you can, please add more tests for the named argument constructor usage of the annotation classes. Those changes don't seem to be covered well by the current tests.

Yoann-TYT:
Thanks for your review @derrabus, I'll see that

phpunit.xml.dist
```diff
@@ -9,7 +9,7 @@
     <php>
         <server name="KERNEL_CLASS" value="Tests\Fixtures\TestKernel" />
         <server name="KERNEL_DIR" value="tests/Fixtures/" />
-        <server name="SYMFONY_PHPUNIT_VERSION" value="7.5" />
+        <server name="SYMFONY_PHPUNIT_VERSION" value="8.5.13" />
```
derrabus, line 12: By the way, do we want/need to pin the patch version here?

```suggestion
        <server name="SYMFONY_PHPUNIT_VERSION" value="8.5" />
```

src/Configuration/Cache.php
Yoann-TYT: I'm agree with you but i'm annoyed because on the documentation, we suggest to use "Etag" and not "etag" for the annotation Cache. 
So, I can transform Etag to etag, but I need to add a information on the documentation to inform that for attributes, we use etag and not Etag. What do you think about that ?

Yoann-TYT:
> Thank you for working on this topic! I have added a few comments.
> 
> If you can, please add more tests for the named argument constructor usage of the annotation classes. Those changes don't seem to be covered well by the current tests.

Yes, I'll add more tests this week-end ;)

src/EventListener/ControllerListener.php
```diff
@@ -60,6 +66,35 @@ public function onKernelController(KernelEvent $event)
...
+                            Template::class
+                        ]
+                    )
+                    ) {
+                        return $value->newInstance();
+                    }
+                }, $attributes);
+            };
+            $classAttributes = $object->getAttributes();
+            $createAttributes($classAttributes);
```
derrabus, line 89:
```suggestion
            $classAttributes = array_map(
                function (\ReflectionAttribute $attribute) {
                    return $attribute->newInstance();
                },
                $object->getAttributes(ConfigurationAnnotation::class, \ReflectionAttribute::IS_INSTANCEOF)
            );
```

src/Configuration/Cache.php
```diff
@@ -101,6 +102,43 @@ class Cache extends ConfigurationAnnotation
      */
     private $staleIfError;
 
+    /**
+     * Cache constructor.
+     * @param int|string $maxage
```
andrew-demb, line 107:
```suggestion
     * @param int|string|null $maxage
```

src/Configuration/Cache.php
```diff
@@ -101,6 +102,43 @@ class Cache extends ConfigurationAnnotation
      */
     private $staleIfError;
 
+    /**
+     * Cache constructor.
```
derrabus, line 106:
```suggestion
```

src/Configuration/Cache.php
```diff
@@ -30,15 +31,15 @@ class Cache extends ConfigurationAnnotation
      * The number of seconds that the response is considered fresh by a private
      * cache like a web browser.
      *
-     * @var int
+     * @var int|string
```
derrabus, line 34: Why do we suddenly allow a string here?

src/Configuration/Entity.php
```diff
@@ -26,4 +27,30 @@ public function setExpr($expr)
 
         $this->setOptions($options);
     }
+
+    /**
+     * Entity constructor.
```
derrabus, line 32:
```suggestion
```

src/Configuration/IsGranted.php
```diff
@@ -50,6 +51,30 @@ class IsGranted extends ConfigurationAnnotation
      */
     private $statusCode;
 
+    /**
+     * IsGranted constructor.
```
derrabus, line 55:
```suggestion
```

src/Configuration/Cache.php
Yoann-TYT: Because it's possible for smaxage, maxage and maxstale attributes to set string with relative time format (1 day, 2 weeks, …)
https://symfony.com/doc/current/bundles/SensioFrameworkExtraBundle/annotations/cache.html

derrabus: Oh, so the `@var` was lying to us the whole time. Got it. :-)

src/Configuration/ParamConverter.php
```diff
@@ -54,6 +55,30 @@ class ParamConverter extends ConfigurationAnnotation
      */
     private $converter;
 
+    /**
+     * ParamConverter constructor.
```
derrabus, line 59:
```suggestion
```

src/Configuration/Security.php
```diff
@@ -43,6 +44,28 @@ class Security extends ConfigurationAnnotation
      */
     protected $message = 'Access denied.';
 
+    /**
+     * Security constructor.
```
derrabus, line 48:
```suggestion
```

src/Configuration/Template.php
```diff
@@ -47,6 +48,30 @@ class Template extends ConfigurationAnnotation
      */
     private $owner = [];
 
+    /**
+     * Template constructor.
```
derrabus, line 52:
```suggestion
```

src/Configuration/Cache.php
derrabus, line 34:
```suggestion
     * @var int|string|null
```

src/Configuration/Cache.php
```diff
@@ -30,15 +31,15 @@ class Cache extends ConfigurationAnnotation
...
      */
     private $maxage;
 
     /**
      * The number of seconds that the response is considered fresh by a public
      * cache like a reverse proxy cache.
      *
-     * @var int
+     * @var int|string
```
derrabus, line 42:
```suggestion
     * @var int|string|null
```

derrabus (changes requested):

fabpot:
Can you rebase on current master (that should help for the tests)

src/Configuration/Cache.php
```diff
@@ -101,6 +102,42 @@ class Cache extends ConfigurationAnnotation
      */
     private $staleIfError;
 
+    /**
+     * @param int|string|null $maxage
+     * @param int|string      $smaxage
+     * @param int|string      $maxstale
+     * @param int|string      $staleWhileRevalidate
+     * @param int|string      $staleIfError
```
derrabus, line 110:
```suggestion
     * @param int|string|null $smaxage
     * @param int|string|null $maxstale
     * @param int|string|null $staleWhileRevalidate
     * @param int|string|null $staleIfError
```

tests/EventListener/ControllerListenerTest.php
```diff
@@ -108,11 +158,34 @@ public function testMultipleParamConverterAnnotationsOnMethod()
         $annotations = $this->request->attributes->get('_converters');
         $this->assertNotNull($annotations);
         $this->assertArrayHasKey(0, $annotations);
-        $this->assertInstanceOf('Sensio\Bundle\FrameworkExtraBundle\Configuration\ParamConverter', $annotations[0]);
+        $this->assertInstanceOf(\Sensio\Bundle\FrameworkExtraBundle\Configuration\ParamConverter::class, $annotations[0]);
```
derrabus, line 161: We should import the classes instead of using FQCNs here.

derrabus (changes requested):
I think, we're on a good path here, thank you so much for your work.

Please remember adding tests that cover using each of the annotation classes as an attribute. So far, only `Chache` is really covered and there's only a basic test for `ParamConverter`.

nlhommet:
Hi,
What is blocking? need help finishing this PR?

Yoann-TYT:
Hi @nlhommet  ! 
Thanks, I just need to add more tests. I'll do that this week-end ! :)

src/Configuration/Cache.php
```diff
@@ -11,12 +11,15 @@
 
 namespace Sensio\Bundle\FrameworkExtraBundle\Configuration;
 
+use Attribute;
```
derrabus, line 14: We don't import classes from the global namespace.

Yoann-TYT: I agree but fabbot.io fail if I don't do that :-/

fabpot: fabbot is wrong :)

fabpot:
Thanks @Yoann-TYT for working on this feature, this is much appreciated.

src/Configuration/Security.php
```diff
@@ -43,6 +44,27 @@ class Security extends ConfigurationAnnotation
      */
     protected $message = 'Access denied.';
 
+    /**
+     * @param array|string $data
+     */
+    public function __construct(
+        $data = [],
+        string $message = null,
```
kevinpapst, line 52: This is a BC break, I opened #733 for more info

Issue #709, opened 2021-01-20
Psr7 is used implicitly and breaks configuration


Wirone:
I wanted to add my 2 cents to #620 because I got the same error after today's update, but I decided to raise new issue.

I use `ekreative/uuid-extra-bundle` which requires `sensio/framework-extra-bundle` which [defines aliases](src/Resources/config/psr7.xml) based on `nyholm/psr7` without requiring that package explicitly (it's in `require-dev`). Today I've updated `sentry/sentry-symfony` to 4.0 (PHP8 support) and in the end I got:

> Uncaught Error: Class "Nyholm\Psr7\Factory\Psr17Factory" not found

I was able to get rid of this error by disabling `psr_message` in `config/packages/sensio_framework_extra.yaml` BUT I took a look at `Sensio\Bundle\FrameworkExtraBundle\DependencyInjection\Configuration` and it seems like PSR7 feature is enabled if `HttpFoundationFactoryInterface` from `symfony/psr-http-message-bridge` exists. In my case this package is required by Sentry package, so basically my application broke down because of implicit requirement for `sensio/framework-extra-bundle` AND `symfony/psr-http-message-bridge`, each required by different packages. Seems like checking for `HttpFoundationFactoryInterface` only is too optimistic and it not guarantee this feature will work.

I think this bundle should do one of those:
- `psr7.xml` should be loaded by default only if all classes used there exist (probably checking for `Nyholm\Psr7\Factory\Psr17Factory` would do the trick)
- explicitly require `nyholm/psr7`, which will solve all required dependencies

Let me know if this will be considered as bug so maybe I can work on it.

fabpot:
See my "fix" in #710

Wirone:
@fabpot yeah, if you want to fix this issue in *that way*, it will work too - it's up to you :wink: I don't know this feature/bundle well enough to help on decision-making level, I just wanted to point that default configuration needs a tweak. Anyway, thanks for very quick reaction :+1:

fabpot:
Fixed in upcoming 6.0

jeroennoten:
Is there a plan/schedule for when 6.0 will be released?

fabpot:
6.0 just released.

jeroennoten:
Awesome! Thanks @fabpot

Issue #620, opened 2019-06-04
upgrade from 5.2.4 to 5.3.1 gives `Error: Class 'Nyholm\Psr7\Factory\Psr17Factory' not found`


rbaarsma:
I've just upgraded from sensio/framework-extra-bundle 5.2.4 to 5.3.1 in my symfony 4.2.9 project.

Now I get this message on several tests:

```
Error: Class 'Nyholm\Psr7\Factory\Psr17Factory' not found

/var/www/application/var/cache/test/ContainerGHVBRZx/getSensioFrameworkExtra_Psr7_HttpMessageFactoryService.php:9
/var/www/application/var/cache/test/ContainerGHVBRZx/srcPublicationService_KernelTestDebugContainer.php:591
/var/www/application/var/cache/test/ContainerGHVBRZx/getSecurity_Authentication_Listener_Oauth2_ApiService.php:12
/var/www/application/var/cache/test/ContainerGHVBRZx/srcPublicationService_KernelTestDebugContainer.php:591
/var/www/application/var/cache/test/ContainerGHVBRZx/getSecurity_Firewall_Map_Context_ApiService.php:13
/var/www/application/vendor/symfony/security/Http/Firewall.php:71
/var/www/application/vendor/symfony/security-bundle/Debug/TraceableFirewallListener.php:33
/var/www/application/vendor/symfony/security/Http/Firewall.php:90
/var/www/application/vendor/symfony/security-bundle/EventListener/FirewallListener.php:48
/var/www/application/vendor/symfony/event-dispatcher/Debug/WrappedListener.php:115
/var/www/application/vendor/symfony/event-dispatcher/EventDispatcher.php:212
/var/www/application/vendor/symfony/event-dispatcher/EventDispatcher.php:44
/var/www/application/vendor/symfony/event-dispatcher/Debug/TraceableEventDispatcher.php:145
/var/www/application/vendor/symfony/http-kernel/HttpKernel.php:126
/var/www/application/vendor/symfony/http-kernel/HttpKernel.php:67
/var/www/application/vendor/symfony/http-kernel/Kernel.php:198
/var/www/application/vendor/symfony/http-kernel/Client.php:68
/var/www/application/vendor/symfony/framework-bundle/Client.php:131
/var/www/application/vendor/symfony/browser-kit/Client.php:405
/var/www/application/tests/functional/Application/Controller/CourseApiControllerTest.php:44
```

I also get this message on other tests

```
[...]

PR #710, opened 2021-01-20
Remove PSR-7 support

fabpot:
People have issues with it (#620, #709, ...) and it does not have to be part of this bundle (I've never used it myself and would never recommend anyone to use it anyway, it adds overhead for no benefits).

If someone wants it, creating a new bundle is the way to go.


nicolas-grekas (approved):

Tobion:
Aren't you adding deprecations for this first?

fabpot:
I'll let someone do the PR if that's something useful.

ajgarlag:
I have released [`ajgarlag/psr-http-message-bundle`](https://github.com/ajgarlag/psr-http-message-bundle). It adds PSR-7 support to Symfony Framework apps.

I hope it can help anyone depending on this removed feature. Any feedback is welcome.

Tobion:
See https://github.com/symfony/psr-http-message-bridge/pull/89 and https://github.com/symfony/recipes/pull/899 for the new solution of this feature.

PR #726, opened 2021-05-06
Edited bug for PHP 8.0

Viktor-Truhanovich:
Attempted to call an undefined method named "isBuiltin" of class "ReflectionUnionType".

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,7 @@
         }
     ],
     "require": {
-        "php": ">=8.0",
+        "php": ">=7.2.5",
         "doctrine/annotations": "^1.0",
         "symfony/config": "^4.4|^5.0",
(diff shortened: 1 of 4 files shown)


composer.json
```diff
@@ -11,7 +11,7 @@
         }
     ],
     "require": {
-        "php": ">=7.2.5",
+        "php": ">=8.0",
```
fabpot, line 14: I don't think we should drop PHP 7 compact. Any way to make the code work for all currently supported versions?

Viktor-Truhanovich: I fixed it

Viktor-Truhanovich (commented):
Edited support of PHP7

Viktor-Truhanovich:
> Edited support of PHP7



> Edited support of PHP7

PR #727, opened 2021-05-07
Added support PHP 8

Viktor-Truhanovich:
Attempted to call an undefined method named "isBuiltin" of class "ReflectionUnionType".

Fixed setting server to route. It fixes tests.

### DIFF
--- a/src/Configuration/Route.php
+++ b/src/Configuration/Route.php
@@ -39,6 +39,16 @@ class Route extends BaseRoute
         return $this->service;
     }
 
+    public function setLocalizedPaths(array $localizedPaths)
+    {
+        if (isset($localizedPaths['service'])) {
+            $this->setService($localizedPaths['service']);
+            unset($localizedPaths['service']);
+        }
+
+        parent::setLocalizedPaths($localizedPaths);
+    }
+
     /**
      * Multiple route annotations are allowed.
      *
--- a/src/EventListener/ParamConverterListener.php
+++ b/src/EventListener/ParamConverterListener.php
@@ -76,7 +76,7 @@ class ParamConverterListener implements EventSubscriberInterface
     {
         foreach ($r->getParameters() as $param) {
             $type = $param->getType();
-            $class = null !== $type && !$type->isBuiltin() ? $type->getName() : null;
+            $class = $this->getParamClassByType($type);
             if (null !== $class && $request instanceof $class) {
                 continue;
             }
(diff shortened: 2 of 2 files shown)


src/Configuration/Route.php
```diff
@@ -39,6 +39,16 @@ public function getService()
         return $this->service;
     }
 
+    public function setLocalizedPaths(array $localizedPaths)
```
fabpot, line 42: This change does not seem to be related.

Viktor-Truhanovich: Yes, not connected. But tests were falling.

fabpot:
Thank you @Viktor-Truhanovich.

src/EventListener/ParamConverterListener.php
```diff
@@ -104,6 +104,28 @@ private function autoConfigure(\ReflectionFunctionAbstract $r, Request $request,
         return $configurations;
     }
 
+    private function getParamClassByType(?\ReflectionType $type): ?string
+    {
+        if (\PHP_VERSION_ID < 80000) {
+            return (null === $type || $type->isBuiltin() || !$type instanceof \ReflectionNamedType) ?
+                null : $type->getName();
+        }
```
derrabus, line 112: I don't think we need this block.

src/EventListener/ParamConverterListener.php
```diff
@@ -104,6 +104,28 @@ private function autoConfigure(\ReflectionFunctionAbstract $r, Request $request,
...
+        foreach ($type instanceof \ReflectionUnionType ? $type->getTypes() : [$type] as $type) {
+            if (null === $type || $type->isBuiltin() || !$type instanceof \ReflectionNamedType) {
+                continue;
+            }
+
+            $class = new \ReflectionClass($type->getName());
+
+            if (!$class->isInterface() && !$class->isAbstract()) {
+                return $type->getName();
+            }
```
derrabus, line 123: Why do you exclude interfaces and abstract classes? This was not the case before your patch. In fact, this change breaks quite a few applications I'm maintaining at the moment.

derrabus (commented):
Sorry, I missed this PR. The new release breaks my apps. I'll work on a fix.

src/EventListener/ParamConverterListener.php
Viktor-Truhanovich: If in the following code we change the order of enumeration of types from Account|EntityInterface to EntityInterface|Account, then without this addition we get an error: Cannot resolve argument $account of "AccountController::read()": Cannot determine controller argument for "AccountController::read()": the $ account argument is type-hinted with the non-existent class or interface:" EntityInterface|Account ".
```
/**
 * @Rest\Route("accounts")
 */
class AccountController extends AbstractFOSRestController implements ApiControllerInterface
{
    /**
     * @Rest\Get("/{id}", requirements={"id"="\d+"})
     *
     * @param Account|EntityInterface $account
     *
     * @return View
     */
    public function read(Account|EntityInterface $account): View
    {
        return $this->accountResponseManager->create($account);
    }
}
```
This addition allows you not to depend on the order of enumerated types, especially since an abstract class and interface cannot be an entity, and a controller action must take an entity

derrabus: > especially since an abstract class and interface cannot be an entity

A Doctrine entity is not the only use-case for a param converter. An application might very well define its own custom param converter that is able to guess a parameter from an interface. This has worked before and we must not break that, especially not with a bugfix release.

I agree that the current implementation does not provide a good algorithm to resolve ambiguous union types. What we've achieved is that `ParamConverterListener` does not error if it comes across a union type. That's great! If you need a better resolution of union types, a PR is more than welcome. But you cannot just throw out support for interfaces just because your app doesn't use them.

Viktor-Truhanovich: Excuse for troubling. Thanks for clarifying

derrabus: No worries. 🙂 Do you plan to work on improving the union type support?

Viktor-Truhanovich: Yes, of course

PR #728, opened 2021-05-17
Don't skip interfaces for auto-conversion

derrabus:
This PR fixes a problem introduced by #727. Before said patch, configuring a param converter for an interface was possible. I've added a test to make sure that this use-case does not break again. In addition to that, I've simplified the code a bit.

### DIFF
--- a/src/EventListener/ParamConverterListener.php
+++ b/src/EventListener/ParamConverterListener.php
@@ -106,19 +106,12 @@ class ParamConverterListener implements EventSubscriberInterface
 
     private function getParamClassByType(?\ReflectionType $type): ?string
     {
-        if (\PHP_VERSION_ID < 80000) {
-            return (null === $type || $type->isBuiltin() || !$type instanceof \ReflectionNamedType) ?
-                null : $type->getName();
+        if (null === $type) {
+            return null;
         }
 
         foreach ($type instanceof \ReflectionUnionType ? $type->getTypes() : [$type] as $type) {
-            if (null === $type || $type->isBuiltin() || !$type instanceof \ReflectionNamedType) {
-                continue;
-            }
-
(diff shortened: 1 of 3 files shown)


src/EventListener/ParamConverterListener.php
```diff
@@ -106,19 +106,12 @@ private function autoConfigure(\ReflectionFunctionAbstract $r, Request $request,
 
     private function getParamClassByType(?\ReflectionType $type): ?string
     {
-        if (\PHP_VERSION_ID < 80000) {
```
derrabus, line 109: 💡 We don't need no special treatment for PHP 7 here. The condition `$type instanceof \ReflectionUnionType` below already does the trick by always evaluating to `false` on PHP 7.

src/EventListener/ParamConverterListener.php
```diff
@@ -106,19 +106,12 @@ private function autoConfigure(\ReflectionFunctionAbstract $r, Request $request,
...
-            return (null === $type || $type->isBuiltin() || !$type instanceof \ReflectionNamedType) ?
-                null : $type->getName();
+        if (null === $type) {
+            return null;
         }
 
         foreach ($type instanceof \ReflectionUnionType ? $type->getTypes() : [$type] as $type) {
-            if (null === $type || $type->isBuiltin() || !$type instanceof \ReflectionNamedType) {
```
derrabus, line 115: 💡 This is redundant. On PHP 7, `$type` is always `ReflectionNamedType` because that's the only implementation of `ReflectionType` there. On PHP 8.0, `ReflectionUnionType::getTypes()` always returns an array of `ReflectionNamedType` instances.

fabpot (approved):

PR #729, opened 2021-05-21
Extend PHP 8.0 attribute support to `ConfigurationInterface`

### LINKED ISSUE #730: Extending PHP 8.0 attribute support to `ConfigurationInterface`
Hi! I created #729 to support cases when we want to create a custom attribute but we don't want to extend `Sensio\Bundle\FrameworkExtraBundle\Configuration\ConfigurationAnnotation`. I believe no BC will occur because of this change and in general it's just an extension of the existing behavior with no impact apart mentioned in the motivation.

Hopefully it can be merged. Thanks!

denisvmedia:
This PR replaces `Sensio\Bundle\FrameworkExtraBundle\Configuration\ConfigurationAnnotation` with `Sensio\Bundle\FrameworkExtraBundle\Configuration\ConfigurationInterface` when processing PHP 8.0 attributes. This allows custom annotations/attributes to directly implement `ConfigurationInterface` rather than extend `ConfigurationAnnotation`.

Closes #730.

### DIFF
--- a/src/EventListener/ControllerListener.php
+++ b/src/EventListener/ControllerListener.php
@@ -13,7 +13,6 @@ namespace Sensio\Bundle\FrameworkExtraBundle\EventListener;
 
 use Doctrine\Common\Annotations\Reader;
 use Doctrine\Persistence\Proxy;
-use Sensio\Bundle\FrameworkExtraBundle\Configuration\ConfigurationAnnotation;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\ConfigurationInterface;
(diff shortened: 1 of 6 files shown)


derrabus:
Can you provide a test case that covers your change?

denisvmedia:
> Can you provide a test case that covers your change?

Sure. Just pushed the tests.

bobvandevijver:
What is the status of this? I would like to see this same change: I do not need the `ConfigurationAnnotation` as I do not need to support the array based constructor parameters as I want to create attributes only.

derrabus (approved):

fabpot:
Thank you @denisvmedia.

PR #731, opened 2021-05-21
Add PHP 8 attribute docs

### LINKED ISSUE #718: Docs are missing support for attribute syntax
As I am currently trying to migrate my codebase from PHP 7.4 to 8.0 while simultaneously migrating Symfony 5.0 to 5.2, I am sometimes struggling with the syntax for attributes (which I prefer over annotations). While it can seem trivial to implement the change, it's still nice to see the expected result.

I was trying to find examples I could analyze, but saw that none of the [official docs for SensioFrameworkExtraBundle](https://symfony.com/doc/current/bundles/SensioFrameworkExtraBundle/index.html) acknowledge that attributes even work (even though the code obviously is adapted).

Is that by choice? Or is the project still waiting for more widespread adoption of the attribute syntax? Or is it just that nobody has had the time to do that?

denisvmedia:
Closes #718

### DIFF
--- a/src/Resources/doc/annotations/cache.rst
+++ b/src/Resources/doc/annotations/cache.rst
@@ -9,40 +9,81 @@ HTTP Expiration Strategies
 
 The ``@Cache`` annotation allows to define HTTP caching::
 
-    use Sensio\Bundle\FrameworkExtraBundle\Configuration\Cache;
+.. configuration-block::
 
-    /**
-     * @Cache(expires="tomorrow", public=true)
-     */
-    public function index()
-    {
-    }
+    .. code-block:: php-annotations
 
-You can also use the annotation on a class to define caching for all actions
-of a controller::
+        use Sensio\Bundle\FrameworkExtraBundle\Configuration\Cache;
 
-    /**
-     * @Cache(expires="tomorrow", public=true)
-     */
-    class BlogController extends Controller
-    {
-    }
-
-When there is a conflict between the class configuration and the method
-configuration, the latter overrides the former::
-
-    /**
-     * @Cache(expires="tomorrow")
-     */
-    class BlogController extends Controller
-    {
         /**
-         * @Cache(expires="+2 days")
+         * @Cache(expires="tomorrow", public=true)
          */
         public function index()
         {
         }
-    }
+
+    .. code-block:: php-attributes
+
+        use Sensio\Bundle\FrameworkExtraBundle\Configuration\Cache;
+
+        #[Cache(expired: 'tomorrow', public: true)]
+        public function index()
+        {
+        }
+
+You can also use the annotation on a class to define caching for all actions
+of a controller::
+
+.. configuration-block::
+
+    .. code-block:: php-annotations
+
+        /**
+         * @Cache(expires="tomorrow", public=true)
+         */
+        class BlogController extends Controller
+        {
+        }
+
+    .. code-block:: php-attributes
+
+        #[Cache(expired: 'tomorrow', public: true)]
+        class BlogController extends Controller
+        {
+        }
+
+When there is a conflict between the class configuration and the method
+configuration, the latter overrides the former::
+
+.. configuration-block::
+
+    .. code-block:: php-annotations
+
+        /**
+         * @Cache(expires="tomorrow")
+         */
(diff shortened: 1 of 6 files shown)


derrabus:
Thank you for your contribution. I find it highly confusing to mix annotations and attributes into the same example. Can we use separate code blocks for them, like in symfony/symfony-docs#14230?

denisvmedia:
OK, I will update the PR.

denisvmedia:
I updated the docs. Hopefully, I did it right, as it was my first time to work with the rst format.

src/Resources/doc/annotations/cache.rst
```diff
@@ -9,40 +9,85 @@ HTTP Expiration Strategies
...
-         * @Cache(expires="+2 days")
+         * @Cache(expires="tomorrow")
          */
-        public function index()
+        class BlogController extends Controller
         {
+            /**
+             * @Cache(expires="+2 days")
+             */
+            #[Cache(expired: '+2 days')]
```
maxhelias, line 70: Maybe a bad copy/past :D
```suggestion
```

src/Resources/doc/annotations/cache.rst
```diff
@@ -9,40 +9,85 @@ HTTP Expiration Strategies
...
-    }
+
+    .. code-block:: php-attributes
+
+        #[Cache(expired: 'tomorrow')]
+        class BlogController extends Controller
+        {
+            /**
+             * @Cache(expires="+2 days")
+             */
```
maxhelias, line 83:
```suggestion
```

src/Resources/doc/annotations/converters.rst
```diff
@@ -15,6 +15,8 @@ they can be injected as controller method arguments::
      * @Route("/blog/{id}")
      * @ParamConverter("post", class="SensioBlogBundle:Post")
      */
+    #[Route('/blog/{id}')]
```
maxhelias, line 18: I think we can/need separate code blocks here too

src/Resources/doc/annotations/routing.rst
```diff
@@ -61,6 +77,28 @@ restrict the HTTP methods of the route:
...
+    -use Sensio\Bundle\FrameworkExtraBundle\Configuration\Route;
+    -use Sensio\Bundle\FrameworkExtraBundle\Configuration\Method;
+    +use Symfony\Component\Routing\Annotation\Route;
+
+    class DefaultController extends Controller
+    {
+        /**
+    -      * @Route("/show/{id}")
+    -      * @Method({"GET", "HEAD"})
+         */
```
maxhelias, line 93:
```suggestion
    -    /**
    -      * @Route("/show/{id}")
    -      * @Method({"GET", "HEAD"})
    -     */
```

maxhelias (commented):
LGTM, just some small change

maxhelias (approved):

src/Resources/doc/annotations/routing.rst
```diff
@@ -29,6 +29,22 @@ annotation, so you only have to update the annotation class namespace:
...
+Alternatively, it can be done with the help of the PHP 8 attribute:
+
+.. code-block:: diff
+
+    -use Sensio\Bundle\FrameworkExtraBundle\Configuration\Route;
+    +use Symfony\Component\Routing\Annotation\Route;
+
+    class DefaultController extends Controller
+    {
+        #[Route('/')]
```
derrabus, line 41: We should render the old route as a Doctrine annotation, shouldn't we?

denisvmedia: I'm not sure about it. If it's so, we just need to remove this block at all. Please advise :)

derrabus: I mean, for the other example for the migration to Symfony's `Route` annotation, this is what you do. But I think, it would be just fine if we ignored attributes for the migration completely.

denisvmedia: @derrabus got you. I've just updated it to make consistent with the other blocks. Could it be done this way?

derrabus (approved):

fabpot:
Thank you @denisvmedia.

PR #734, opened 2021-05-31
fix empty message for security annotation/attribute

### LINKED ISSUE #733: BC break in 6.1.0
Hello!

When upgrading a project to the latest release, I ran into a BC break which is a blocker right now for me.

It was introduced by @Yoann-TYT in #707 when adding the PHP8 Attributes support. 

Reason is this line:
https://github.com/sensiolabs/SensioFrameworkExtraBundle/blob/5e7f568160ecae3c16b14264bbc57d2a4a854df9/src/Configuration/Security.php#L62

which overwrites the default `protected $message = 'Access denied.';` with `string $message = null` from the constructor.
This triggers a "strange" exception message. Instead of seeing `Access denied` I now see `Expression "is_granted('my_acl')" denied access.` caused by the SecurityListener:
https://github.com/sensiolabs/SensioFrameworkExtraBundle/blob/5e7f568160ecae3c16b14264bbc57d2a4a854df9/src/EventListener/SecurityListener.php#L79

because the $message is now null instead of `Access denied.` in the $configuration object.

I guess the same exception message is triggered when using Attributes. In both cases the default message `Access denied` should be used. Right now, I am using the Exception message in an API response and I definitely don't want to leak the Security expression which raised the Exception.

kevinpapst:
Fixes #733

### DIFF
--- a/src/Configuration/Security.php
+++ b/src/Configuration/Security.php
@@ -59,7 +59,7 @@ class Security extends ConfigurationAnnotation
             $values = $data;
         }
 
-        $values['message'] = $values['message'] ?? $message;
+        $values['message'] = $values['message'] ?? $message ?? $this->message;
         $values['statusCode'] = $values['statusCode'] ?? $statusCode;
 
         parent::__construct($values);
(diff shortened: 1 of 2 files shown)


fabpot (approved):

fabpot:
Thank you @kevinpapst.

PR #737, opened 2021-06-14
Update index.rst

BooleanType:
`#[Cache(expired: 'tomorrow')]` attribute is omitted in code block. I think, it should be added for compatibility.


fabpot:
Thank you @BooleanType.

PR #740, opened 2021-07-02
Add IS_REPEATABLE on the Security attribute

Neutron-Pro:
This addition allows to manage several different security cases.

### DIFF
--- a/src/Configuration/Security.php
+++ b/src/Configuration/Security.php
@@ -17,7 +17,7 @@ namespace Sensio\Bundle\FrameworkExtraBundle\Configuration;
  * @author Fabien Potencier <fabien@symfony.com>
  * @Annotation
  */
-#[\Attribute(\Attribute::TARGET_CLASS | \Attribute::TARGET_METHOD)]
+#[\Attribute(\Attribute::TARGET_CLASS | \Attribute::TARGET_METHOD | \Attribute::IS_REPEATABLE)]
 class Security extends ConfigurationAnnotation
 {
     /**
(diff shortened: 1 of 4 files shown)


derrabus:
Good catch! Can you please add a test?

Neutron-Pro:
I would like to make one but actually I don't know how to make a unit test. However, I have tested it on my side and it works very well since it is based on the principle of annotations.

```php
#[Route('/banks/{id<%uuid_pattern%>}', methods: ['DELETE'])]
#[Security('bank.getDeletedAt() === null', 'Bank not found', 404)]
#[Security('is_granted(\'ROLE_BANK_DELETE\') or bank.getUser() === user')]
#[Security(
    'request.request.get(\'delete\', \'\') === bank.getName()',
    'You must insert the name of the bank correctly spelled !',
    401
)]
class DeleteController {
  // ...
}
```

If you have an example of unit test to implement it, I do it without problem. :)

derrabus:
I trust that your change is correct. I'd like to avoid that we break your use-case in the future. 🙂 

I'm on a phone right now, I'll try to find some time later to help you. In the meantime, you could look at existing tests. Maybe you'll find some inspiration there.

Neutron-Pro:
Yes I understand, I wanted to take inspiration from the IsGranted which already accepts repetition but it doesn't exist, or I didn't find it (which is very likely). I'm checking what already exists, if I find it I'll add it, if not I don't say no to a little help :)

Neutron-Pro:
I probably found it though, when I run the unit tests (without having added my tests) I get this error:

```
InvalidArgumentException: Invalid environment provided to "Mock_Kernel_6e3d9150": the environment cannot be empty.
```
or
```
Failed asserting that exception message 'Invalid environment provided to "Mock_Kernel_6e3d9150": the environment cannot be empty.' contains 'Multiple "cache" annotations are not allowed'.
```

Have I forgotten something ?

Neutron-Pro:
I added the unit tests and everything passes for 7.2. However, as I said before, higher versions don't pass and I don't think it's specifically me.

fabpot:
See #783

PR #742, opened 2021-07-18
Allow Symfony 6

Kocal:
Some tests are failing, but that was already the case before this PR (eg: in https://github.com/sensiolabs/SensioFrameworkExtraBundle/runs/3084739718?check_suite_focus=true).
Unfortunately I don't know how to fix them.

Thanks!


tests/Fixtures/config/config_sf5.yml
```diff
@@ -0,0 +1,3 @@
+security:
+  password_hashers:
+    Symfony\Component\Security\Core\User\User: plaintext
```
derrabus, line 3: Please use four spaces for indentation

Kocal: Fixed, thanks!

Kocal: Fixed, thanks!

Nyholm (approved):
Thank you. 

I agree with this change. I would be happy to see it merged.

fabpot:
Thank you @Kocal.

PR #743, opened 2021-07-20
Fix comment indent

Seb33300:


fabpot:
Good catch, thanks @Seb33300.

PR #720, opened 2021-03-25
bug #699 Accept null arguments as subjects for IsGranted annotation

nfragnet:
This PR solves #699 by managing nullable subjects or array subjects with null values on existing keys.

This may also replace a previous PR #703

PR #749, opened 2021-08-06
Allow null subject for @IsGranted (fixes #747)

### LINKED ISSUE #747: @IsGranted annotation fails for nullable parameter
Nullable parameters with IsGranted cause it to fail with `Could not find the subject "[subject]" for the @IsGranted annotation. Try adding a "$[subject]" argument to your controller method.` when the parameter is null.

AndrolGenhald:


fabpot:
Fixed by #720

PR #750, opened 2021-08-12
Doc, fix wrong array notation in attribute

ZHB:


fabpot:
Good catch, thanks @ZHB.

PR #756, opened 2021-11-14
Fix typo in Cache attribute docs

t-richard:
The property is named `expires`, not `expired`

See https://github.com/sensiolabs/SensioFrameworkExtraBundle/blob/master/src/Configuration/Cache.php#L28


derrabus (approved):

fabpot (approved):

fabpot:
Thank you @t-richard.

Issue #758, opened 2021-12-03
Error in auto scripts when upgrade to Symfony 6.0


hantsy:
Change the Symfony version to 6.0.*, and run `composer update`, and got the following errors in the console.

```
Script cache:clear returned with error code 1
!!
!!  In XmlFileLoader.php line 407:
!!  

!!    Unable to parse file "D:\hantsylabs\symfony5-sample\rest-sample\vendor\sensio\framewo
r
!!    k-extra-bundle\src\DependencyInjection/../Resources/config\routing.xml": [ERROR 1868]

!!    Element '{http://symfony.com/schema/dic/services}deprecated': The attribute 'package'

!!    is required but missing. (in file:/D:/hantsylabs/symfony5-sample/rest-sample/ - line 
1
!!    1, column 0)

!!    [ERROR 1868] Element '{http://symfony.com/schema/dic/services}deprecated': The attrib
u
!!    te 'version' is required but missing. (in file:/D:/hantsylabs/symfony5-sample/rest-sa
m
!!    ple/ - line 11, column 0)

!!    [ERROR 1868] Element '{http://symfony.com/schema/dic/services}deprecated': The attrib
u
!!    te 'package' is required but missing. (in file:/D:/hantsylabs/symfony5-sample/rest-sa
m
!!    ple/ - line 18, column 0)

!!    [ERROR 1868] Element '{http://symfony.com/schema/dic/services}deprecated': The attrib
u
!!    te 'version' is required but missing. (in file:/D:/hantsylabs/symfony5-sample/rest-sa
m
!!    ple/ - line 18, column 0)

!!    [ERROR 1868] Element '{http://symfony.com/schema/dic/services}deprecated': The attrib
u
!!    te 'package' is required but missing. (in file:/D:/hantsylabs/symfony5-sample/rest-sa
m
!!    ple/ - line 25, column 0)

!!    [ERROR 1868] Element '{http://symfony.com/schema/dic/services}deprecated': The attrib
u
!!    te 'version' is required but missing. (in file:/D:/hantsylabs/symfony5-sample/rest-sa
m
!!    ple/ - line 25, column 0)

!!  

!!
!!  In XmlUtils.php line 95:
!!  

!!    [ERROR 1868] Element '{http://symfony.com/schema/dic/services}deprecated': The attrib
u
!!    te 'package' is required but missing. (in file:/D:/hantsylabs/symfony5-sample/rest-sa
m
[...]

Issue #721, opened 2021-04-09
Upgrade from v5.6.1 to v6.x.x


skekec1:
Hi everyone,

i would like to know whether there are some breaking changes in v6 that requires some refactoring after upgrade. I was not able to find a Migration / Upgrade section anywhere.

Thanks in advance!

Selcuk

hantsy:
I got exception when upgrading to Symfony v6.0.0. 

When I ran the composer update command

```bash 
!
!!  In XmlFileLoader.php line 407:
!!  

!!    Unable to parse file "D:\hantsylabs\symfony5-sample\rest-sample\vendor\sensio\framework-extra-bundle\src\DependencyInjection/../Resources/config\routing.xml": [ERROR 1868] Elemen
t
!!     '{http://symfony.com/schema/dic/services}deprecated': The attribute 'package' is required but missing. (in file:/D:/hantsylabs/symfony5-sample/rest-sample/ - line 11, column 0) 

!!    [ERROR 1868] Element '{http://symfony.com/schema/dic/services}deprecated': The attribute 'version' is required but missing. (in file:/D:/hantsylabs/symfony5-sample/rest-sample/ -

!!    line 11, column 0)

!!    [ERROR 1868] Element '{http://symfony.com/schema/dic/services}deprecated': The attribute 'package' is required but missing. (in file:/D:/hantsylabs/symfony5-sample/rest-sample/ -

!!    line 18, column 0)

!!    [ERROR 1868] Element '{http://symfony.com/schema/dic/services}deprecated': The attribute 'version' is required but missing. (in file:/D:/hantsylabs/symfony5-sample/rest-sample/ -

!!    line 18, column 0)

!!    [ERROR 1868] Element '{http://symfony.com/schema/dic/services}deprecated': The attribute 'package' is required but missing. (in file:/D:/hantsylabs/symfony5-sample/rest-sample/ -

!!    line 25, column 0)

!!    [ERROR 1868] Element '{http://symfony.com/schema/dic/services}deprecated': The attribute 'version' is required but missing. (in file:/D:/hantsylabs/symfony5-sample/rest-sample/ -

!!    line 25, column 0)

!!  

```

skekec1:
Hey,

thank you for the feedback. Seems that we will not get any feedback to this topic from the maintainer. I will check it also and will provide you more information if i can figure out the problem.

derrabus:
[...]

PR #759, opened 2021-12-17
Fix Symfony 6 BC break with required attributes in routing.xml

TomasVotruba:
Closes https://github.com/sensiolabs/SensioFrameworkExtraBundle/issues/758

Closes https://github.com/sensiolabs/SensioFrameworkExtraBundle/issues/721

### DIFF
--- a/src/Resources/config/routing.xml
+++ b/src/Resources/config/routing.xml
@@ -8,21 +8,21 @@
         <service id="sensio_framework_extra.routing.loader.annot_class" class="Sensio\Bundle\FrameworkExtraBundle\Routing\AnnotatedRouteControllerLoader">
             <tag name="routing.loader" />
             <argument type="service" id="annotation_reader" />
-            <deprecated>The "%service_id%" service is deprecated since version 5.2</deprecated>
+            <deprecated package="sensio/framework-extra-bundle" version="5.2">The "%service_id%" service is deprecated since version 5.2</deprecated>
         </service>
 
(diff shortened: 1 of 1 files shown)


fabpot:
Thank you @TomasVotruba.

sprain:
With this change I now get the following error in a Symfony 4.4 project:

```
[Symfony\Component\DependencyInjection\Exception\InvalidArgumentException]                                                                                                                                                                                                                                          
Unable to parse file "/var/www/html/vendor/sensio/framework-extra-bundle/src/DependencyInjection/../Resources/config/routing.xml":
[ERROR 1827] Element '{http://symfony.com/schema/dic/services}deprecated', attribute 'package': The attribute 'package' is not allowed.
 ```
  
 I am trying to find the root cause, but maybe it's obvious to you and you can give me a hint?

TomasVotruba:
This bundle should be probably limited in composer.json to Symfony 5.4 and 6 only. Otherwise there will be always one major Symfony version unhappy about the config :)

derrabus:
> This bundle should be probably limited in composer.json to Symfony 5.4 and 6 only.

Maybe, but right now it isn't. We should make the bundle compatible with 4.4 again.

fabpot:
Reverted and new version released.

fabpot:
Is anyone willing to work on another fix that makes the code work for all versions? As the `routing.xml` file is deprecated anyway, what about having 2 XML files: one compatible with 4.4 and another one for 5+? And we load the right one in SensioFrameworkExtraExtension.

phiphou:
> Reverted and new version released.

I don't understand why you revert these changes, they were for me the only way to get rid of the error message...

PR #761, opened 2021-12-22
Fix Symfony 6 BC break with required attributes in routing.xml

biozshock:
Followup for #759


fabpot:
Thank you @biozshock.

PR #763, opened 2022-01-06
Make the SecurityListener compatible with new Authenticator for #736

### LINKED ISSUE #736: Security is_granted annotation not working with new PUBLIC_ACCESS role
The user with role `PUBLIC_ACCESS` cannot access a route annotated with `@Security("is_granted('something')")`.

Extra informations :
1) It does not even enter the Voter
2) Testing inside a controller the `$this->isGranted('something')` is correctly working

Alsatian67:
Fixes #736 

Symfony Authentication system has changed between version 5.2 and 5.4.
The security bundle uses a config option 'enable_authenticator_manager' set to true if the new system is used.

### DIFF
--- a/src/DependencyInjection/Compiler/EnableAuthenticatorManagerPass.php
+++ b/src/DependencyInjection/Compiler/EnableAuthenticatorManagerPass.php
@@ -1,0 +1,42 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sensio\Bundle\FrameworkExtraBundle\DependencyInjection\Compiler;
+
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
+/**
+ * Optimizes the container by removing unneeded listeners.
+ *
+ * @author Fabien Potencier <fabien@symfony.com>
+ */
+class EnableAuthenticatorManagerPass implements CompilerPassInterface
+{
+    public function process(ContainerBuilder $container)
+    {          
+        if(!$container->has('sensio_framework_extra.security.listener')){
+            return;
+        }
+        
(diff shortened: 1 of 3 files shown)


src/DependencyInjection/Compiler/EnableAuthenticatorManagerPass.php
```diff
@@ -0,0 +1,42 @@
...
+    {          
+        if(!$container->has('sensio_framework_extra.security.listener')){
+            return;
+        }
+        
+        if(!$container->hasExtension('security')){
+            return;
+        }
+        
+        $securityConfig = $container->getExtensionConfig('security')[0];
```
Alsatian67, line 34: Not sure about the [0] here ? I didn't find why getExtensionConfig() returns an array of arrays.
Please review.

derrabus: I'm nore sure we should parse the extension config to fetch this piece of information.

@wouterj Can you have a look please?

Alsatian67:
Is this repo till alive ?

fabpot:
I was waiting for @wouterj, to be honest. Let's see if he has time in the next few days.

src/DependencyInjection/Compiler/EnableAuthenticatorManagerPass.php
wouterj: Yes, I don't think it's a good idea to depend on the SecurityBundle config here.

If I'm correct, we can completely remove the if statement on SecurityListener:35 and always use the variables from line 95-104.
This means that users will have to take care to update their expression to take null tokens into account. But that's only when they explicitly decided to switch to the new system. In Symfony core, we have not considered this a BC break. But I'll let @fabpot decide about that for this bundle.

wouterj:
Sorry, this was so long buried in my notifications that I didn't know this was blocking on me. I've shared my idea now :)

src/DependencyInjection/Compiler/EnableAuthenticatorManagerPass.php
Alsatian67: I agree with your idea, but I was wondering how important the if statement was.

fabpot:
See #783

PR #766, opened 2022-02-02
[Fix] PHP 81 deprecated implicit conversion from float to int

Yoann-TYT:
Hi everybody ! 

This PR fix a deprecated with PHP 8.1

`Deprecated: Implicit conversion from float 1643798519.685619 to int loses precision`

Thanks for your review !

### DIFF
--- a/src/EventListener/HttpCacheListener.php
+++ b/src/EventListener/HttpCacheListener.php
@@ -190,7 +190,7 @@ class HttpCacheListener implements EventSubscriberInterface
         if (!is_numeric($time)) {
             $now = microtime(true);
 
-            $time = ceil(strtotime($time, $now) - $now);
+            $time = ceil(strtotime($time, (int) $now) - $now);
         }
 
         return $time;


src/EventListener/HttpCacheListener.php
```diff
@@ -190,7 +190,7 @@ private function convertToSecondsIfNeeded($time)
         if (!is_numeric($time)) {
             $now = microtime(true);
 
-            $time = ceil(strtotime($time, $now) - $now);
+            $time = ceil(strtotime($time, round($now)) - $now);
```
derrabus, line 193: Let's not change the logic here. In fact one of the failing tests seems related to your change.

```suggestion
            $time = ceil(strtotime($time, (int) $now) - $now);
```

Yoann-TYT:
Thanks @derrabus ! 😁

tjveldhuizen:
@fabpot Is a release containing this improvement to be expected somewhere in the near future? It would make my deprecation logs a lot clearer, because now every line in the log is unique since it contains the current timestamp.

fabpot:
Released

PR #768, opened 2022-02-24
Enum param converter

dreadnip:
Figured it would be a nice QoL improvement to accept and automatically transform enum values passed as parameters.

### DIFF
--- a/src/Request/ParamConverter/EnumParamConverter.php
+++ b/src/Request/ParamConverter/EnumParamConverter.php
@@ -1,0 +1,53 @@
+<?php
+
+/*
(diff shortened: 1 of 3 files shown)


derrabus:
See symfony/symfony#44831

dreadnip:
Ah I see. Guess I'll use this param converter as a temporary solution in my projects until the 6.1 release drops. Thanks :slightly_smiling_face:

derrabus:
You might as well copy the `BackedEnumValueResolver` class from the 6.1 branch into your codebase for the time being. And if it does not work, open a bug. This could help us stabilizing the feature. 🙂

PR #779, opened 2022-10-29
Check _cache type for compatibility with Symfony 6.2

GromNaN:
The new `#[Cache()]` attribute that is added to Symfony 6.2 (https://github.com/symfony/symfony/pull/46880) is stored in the same request attribute `_cache`.

Before this change, if the bundle is enabled and Symfony's Cache attribute is used, we get an error.

```
Call to a member function getSMaxAge() on array
```

This bugfix will ease migration when simultaneously usage of FrameworkExtraBundle and Symfony attributes is necessary.

### DIFF
--- a/src/EventListener/HttpCacheListener.php
+++ b/src/EventListener/HttpCacheListener.php
@@ -11,6 +11,7 @@
 
 namespace Sensio\Bundle\FrameworkExtraBundle\EventListener;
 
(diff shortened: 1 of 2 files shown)


derrabus (approved):

fabpot:
Thank you @GromNaN.

PR #781, opened 2022-11-10
Add some native return types to prevent deprecation hints

dmitryuk:
Fixes several deprecation messages:
```
Method "Symfony\Component\EventDispatcher\EventSubscriberInterface::getSubscribedEvents()" might add "array" as a native return type declaration in the future. Do the same in implementation "Sensio\Bundle\FrameworkExtraBundle\EventListener\XXXListener" now to avoid errors or add an explicit @return annotation to suppress this message.

Method "Symfony\Component\DependencyInjection\Extension\ExtensionInterface::getNamespace()" might add "string" as a native return type declaration in the future. Do the same in implementation "Sensio\Bundle\FrameworkExtraBundle\DependencyInjection\SensioFrameworkExtraExtension" now to avoid errors or add an explicit @return annotation to suppress this message.
```

It will not affect negative for all Symfony and  PHP versions currently supported

### DIFF
--- a/src/DependencyInjection/SensioFrameworkExtraExtension.php
+++ b/src/DependencyInjection/SensioFrameworkExtraExtension.php
@@ -122,10 +122,7 @@ class SensioFrameworkExtraExtension extends Extension
         return __DIR__.'/../Resources/config/schema';
     }
 
-    /**
-     * @return string
-     */
-    public function getNamespace()
+    public function getNamespace(): string
     {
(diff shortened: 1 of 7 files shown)


derrabus (changes requested):
Those are breaking changes that we cannot ship in a minor or bugfix release.

Furthermore, we're already doing what the error message asks us to: We have explicit `@return`annotations.

